PR #15, opened 2018-10-28
New order

hannesm:
rewrite of basically everything. allows startup scripts and tls-less deployments. single unified command format :)


app/vmmc_bistro.ml
```diff
@@ -0,0 +1,256 @@
...
+    in
+    let issuer = X509.subject cert in
+    X509.CA.sign csr ~valid_from ~valid_until ~extensions key issuer
+  in
+  let certificates = `Single ([ mycert ; cert ], tmpkey) in
+  X509_lwt.authenticator (`Ca_file ca) >>= fun authenticator ->
+  Lwt_unix.gethostbyname host >>= fun host_entry ->
+  let host_inet_addr = Array.get host_entry.Lwt_unix.h_addr_list 0 in
+  let fd = Lwt_unix.socket host_entry.Lwt_unix.h_addrtype Lwt_unix.SOCK_STREAM 0 in
+  Lwt_unix.connect fd (Lwt_unix.ADDR_INET (host_inet_addr, port)) >>= fun _ ->
```
cfcs: would be easier to read if it was `>>= fun () ->` imho :)

src/vmm_unix.ml
```diff
@@ -0,0 +1,180 @@
...
+    let prefix = "vmmtap" in
+    let rec find_n x =
+      let nam = prefix ^ string_of_int x in
+      match Bos.OS.Cmd.run Bos.Cmd.(v "ifconfig" % nam) with
+      | Error _ -> nam
+      | Ok _ -> find_n (succ x)
+    in
+    let tap = find_n 0 in
+    Bos.OS.Cmd.run Bos.Cmd.(v "ip" % "tuntap" % "add" % "mode" % "tap" % tap) >>= fun () ->
+    Bos.OS.Cmd.run Bos.Cmd.(v "brctl" % "addif" % bridge % tap) >>= fun () ->
```
cfcs: I think you can use the `iproute2` command
```ocaml
v "ip" % "link" % "set" % tap % "master" % bridge
```
instead of the call to `brctl` (which is part of `bridge-utils` that is not included in debian by default).
May have to additionally
```ocaml
v "ip" % "link" % "set" % tap % "up"
```
but not sure if that's the case here.

Not sure how to best avoid the race condition, but I guess the `vmmtap` namespacing here works for now.

hannesm:
thanks for your comments @cfcs, I pushed c669be8 which address most of your comments.
- in respect to using `ip` instead of `brctl` -- I couldn't get `ip` to attach tap devices to a bridge, but needed `brctl` for this. is this some newer `ip` command since when this works (and if so, do you know since when and whether this is the same as in ubuntu/debian stable?)
- in respect to the `packaging/` subdirectory: this is wip from @sg2342, not sure whether we should quote all the variables (shellcheck.net says yes) ;)
- avoiding mutable bytes: I disagree this being an issue here, since the incoming data is a cstruct (already mutable, yuck), but the bytes value is only local and does not escape

cfcs:
re: `ip`: I have no idea, it works for me, and I think it always has worked that way. I've always used `ifconfig` and `tunctl` so I have no idea how this new fancy stuff works.

Re: packaging scripts: I don't see the point of quoting half of them at least - either we care about word expansion and meta characters in the CWD, or we don't :)

PR #17, opened 2018-11-08
Freebsd system services

sg2342:


hannesm:
oh, thanks! i came across the issue of unset flags breaking the rc scripts (and have some workarounds deployed).

PR #20, opened 2019-02-06
Linux systemd

hannesm:
mainly to remind myself to look into that soon, thanks for the patches -- @cfcs do your changes result in a working albatross? or do you still see locking issues?


hannesm:
the proposed changes (tmpdir/dbdir and systemd scripts) are part of #25, happy to adjust / review there (or test for starters), if it works we should merge #25

PR #23, opened 2019-10-02
changes for solo5 0.6

hannesm:
this is a breaking change in the wire protocol

the unikernel_config is reordered slightly, and block devices is a sequence instead of a utf8-string option


hannesm:
revised and extended:
- wire version is bumped (still need to figure some things out)
- deterministic mac addresses
- metrics from albatross* via metrics library
- restart on failure
- x509 API upgrade to 0.8.0

hannesm:
before this is good to go, I should:
- work a bit harder to be able to read V3 logs (and state file?) -- esp `Unikernel_start` changed (as well as `type config`
- test on linux (maybe with spt, see #20 -- but use solo5-elftool query-abi to figure out which tender to use)

hannesm:
fwiw, the compatibility with old data was restored in 47fef43, including documentation in vmm_asn on how to be careful in the future to avoid writing data that is not readable again (or, which parts not to change). i guess that means this can be fed back into master.

PR #24, opened 2019-11-11
Versioning

hannesm:
i think it is wiser to not embed version in each binary separately. and `--version` prints out the protocol version as well now!

also, provision_ca was buggy: it did just use the requested version, and not check anything. now, all to_cert_extension use the current(!) version.

daemons are reading a wire and reply with the same protocol version. on version mismatches a warning is printed (vmm_lwt / vmm_tls_lwt / vmm_provision). this means a best-effort supporting earlier albatross clients is now there! :)

### DIFF
--- a/client/albatross_client_bistro.ml
+++ b/client/albatross_client_bistro.ml
@@ -9,7 +9,7 @@ let read fd =
   let rec loop () =
     Vmm_tls_lwt.read_tls fd >>= function
     | Error `Eof ->
-      Logs.warn (fun m -> m "eof from server");
+      Logs.debug (fun m -> m "eof from server");
       Lwt.return (Ok ())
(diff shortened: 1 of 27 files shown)


hannesm:
i re-reviewed this and it looks fine! still, it is kind of a lie that AV3 clients will enjoy talking to AV4 servers (though most should be fine). the versioning applied in this PR is for sure shorter and much better than the previous attempt.

the rules being:
- when initiating a connection, use the highest version (this is up for debate, we could integrate version negotiation as in tls -- i.e. client sends their supported versions, server picks one -- but that feels overly complex for now. a command-line parameter for the client to use an earlier version would be sufficient in the next update cycle)
- when replying to a request, use the version used in the request

PR #25, opened 2019-11-21
Linux support

hannesm:
- introduce tmpdir and dbdir (global mutable state, to-be-set by command-line argument of the respective services, sane defaults for FreeBSD and Linux)
- integrate systemd service files (adapted from #20), including some TODO questions (since I lack understanding of systemd)
- move FreeBSD packaging to `packaging/FreeBSD`, populate `packaging/Linux` with the systemd services above


hannesm:
this has been around for long enough, I don't think there's any harm to merge it. :)1

PR #27, opened 2020-03-30
stats: instead of executing the sysctl kinfo_proc twice (for retrieving kinfo_mem and rusage), only execute it once

hannesm:
this is fully backwards-compatible. I intended to integrate statistics for linux (reading `/proc/<pid>/stat`), and by looking at the available information, there'll be some incompatible changes. The main idea is to unify "rusage" and "kinfo_mem", and remove some of its elements (which haven't shown to be useful, or are not available on Linux).

I realise that on Linux the `/proc/<pid>/status` has more information (such as context switches), but is much harder to read and parse - still may be worth using that (but that doesn't contain the system/user time, neither startup time).

The main things of interest are:
- system time used
- user time used
- time since startup (i.e. uptime)
- virtual size (i.e. maximum amount of memory this process may use)
- current RSS (in pages (that's how it is available)) [maybe max rss as well]
- number of swaps
- page faults & reclaims
- block input/output (couldn't find this on Linux)

### DIFF
--- a/stats/albatross_stat_client.ml
+++ b/stats/albatross_stat_client.ml
@@ -4,9 +4,8 @@ open Albatross_stats_pure
 let old_rt = ref 0L
 
 let timer pid vmmapi interval =
-  let rusage = sysctl_rusage pid in
+  let rusage, kinfo_mem = sysctl_kinfo_proc pid in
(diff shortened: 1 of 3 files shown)


hannesm:
merged manually into master

PR #30, opened 2020-04-21
adapt to x509 0.11.0 API changes

hannesm:
had these locally since a while (//cc @dinosaure who may be interested)


dinosaure:
so fast 😊

dinosaure:
Note that even if I pin `x509` with mirleft/ocaml-x509#133, this branch still compiles.

PR #33, opened 2020-05-12
Minor updates to albatrossd

MagnusS:
Make `albatrossd` retry socket connections while waiting for log/console/stats daemons to start and return socket path in the error messages - this should allow the daemons to be started in any order. Also adds an explicit `--disable-stats`  option as it will now retry the socket connection indefinitely if this is not specified (perhaps this should be `--enable-stats` instead though).

### DIFF
--- a/command-line/albatross_cli.ml
+++ b/command-line/albatross_cli.ml
@@ -353,9 +353,13 @@ let exits =
     remote_command_failed ::
   Term.default_exits
 
-let disable_stats =
-  let doc = "Don't try to connect to albatross-stats to report statistics" in
-  Arg.(value & flag & info [ "disable-stats" ] ~doc)
+let enable_stats =
+  let doc = "Connect to albatross-stats to report statistics" in
+  Arg.(value & flag & info [ "enable-stats" ] ~doc)
+
+let retry_connections =
+  let doc = "Number of retries when connecting to other daemons (log, console, stats etc). 0 aborts after one failure, -1 is unlimited retries." in
+  Arg.(value & opt int 3 & info [ "retry-connections" ] ~doc)
 
 let auth_exits =
   [ Term.exit_info ~doc:"on local authentication failure \
--- a/command-line/albatross_cli.ml
+++ b/command-line/albatross_cli.ml
@@ -353,6 +353,10 @@ let exits =
     remote_command_failed ::
   Term.default_exits
 
+let disable_stats =
+  let doc = "Don't try to connect to albatross-stats to report statistics" in
+  Arg.(value & flag & info [ "disable-stats" ] ~doc)
+
 let auth_exits =
   [ Term.exit_info ~doc:"on local authentication failure \
                          (certificate not accepted by remote)"
--- a/daemon/albatrossd.ml
+++ b/daemon/albatrossd.ml
@@ -135,7 +135,7 @@ let write_reply name fd txt (hdr, cmd) =
 
 let m = conn_metrics "unix"
 
-let jump _ influx tmpdir dbdir disable_stats =
+let jump _ influx tmpdir dbdir retries enable_stats =
   Sys.(set_signal sigpipe Signal_ignore);
   Albatross_cli.set_tmpdir tmpdir;
   Albatross_cli.set_dbdir dbdir;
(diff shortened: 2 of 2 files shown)


command-line/albatross_cli.ml
```diff
@@ -353,6 +353,10 @@ let exits =
     remote_command_failed ::
   Term.default_exits
 
+let disable_stats =
+  let doc = "Don't try to connect to albatross-stats to report statistics" in
+  Arg.(value & flag & info [ "disable-stats" ] ~doc)
```
hannesm, line 358: I like this, but as you comment, an `enable-stats` would be even better. :)

daemon/albatrossd.ml
```diff
@@ -145,12 +145,23 @@ let jump _ influx tmpdir dbdir =
...
+      (let rec unix_connect ~retries s =
+         Vmm_lwt.connect Lwt_unix.PF_UNIX (Lwt_unix.ADDR_UNIX (socket_path s)) >>= fun x ->
+         (match x with
+         | Some x -> Lwt.return (Some x)
+         | None -> if (retries-1 != 0) then begin
+                     Logs.err (fun m -> m "unable to connect to %s, retrying in 5 seconds" (socket_path s));
+                     Lwt_unix.sleep 5.0 >>= fun () ->
+                     unix_connect ~retries:(retries-1) s
+                   end else
+                     Lwt.return_none)
```
hannesm, line 157: I'm not too convinced of this approach. At the moment, albatrossd requires the console and log daemon to be executed before albatrossd is started. Once either of them disappears (and albatrossd notices), albatrossd gives up and terminates.

With this change, albatrossd could idle not starting up due to issues in the other daemons. Currently, the order is enforced by the service scripts + rc scripts.

OTOH, it'd be nice to be able to start and stop and restart the daemons completely independently (esp. log and stats -- console may be a bit tricky (if it's reading from any fifos, it is not clear to me whether it can be safely stop (close() the fifo) and started again (open() all fifos)).

MagnusS: Yes, if console or log fails or are slow to start `albatrossd` would wait and report which socket it was waiting for -- currently indefinitely, but I guess there could be a `--retries=` with a lower default if you want it to fail. I just noticed that `albatrossd` also opens its own socket first, which isn't ideal as it's not functional yet (but it looks like that could be moved). I don't mind closing this if you'd prefer to keep the solution as it is for now though -- I mostly wrote this to make my own testing easier :)

hannesm: After sleeping over it, I'd be fine with:
- opening the server socket only after the other connections are up and running (well spotted, thanks)
- accepting an optional `retry-connections` command-line argument to use for the connect-on-startup logic. if it is 0, no retry should be done. its default should be some small number (maybe 3)
- an `--enabled-stats` command-line argument (your `--disable-stats`, but inverted) to control whether a connection to the stats is attempted (if `--enable-stats` is used, and the connection to the stats socket fails, albatrossd should terminate)

MagnusS: Ok, sounds good -- I'll update the PR with these changes (hopefully later today)

MagnusS:
@hannesm I think these latest commits should address your comments. New parameter `--retry-connections` is 3 by default (-1 is unlimited retries), server socket is opened last and stats is enabled with a `--enable-stats` parameter

daemon/albatrossd.ml
```diff
@@ -158,14 +158,22 @@ let jump _ influx tmpdir dbdir disable_stats =
...
+       (unix_connect ~retries `Log >|= function
          | None -> invalid_arg "cannot connect to log socket"
          | Some l -> l) >>= fun l ->
 
-       (unix_connect ~retries:(-1) `Console >|= function
+       (unix_connect ~retries `Console >|= function
          | None -> invalid_arg "cannot connect to console socket"
          | Some c -> c) >>= fun c ->
 
+       (if (enable_stats) then
```
hannesm, line 169:
```suggestion
       (if enable_stats then
```

daemon/albatrossd.ml
```diff
@@ -149,7 +149,7 @@ let jump _ influx tmpdir dbdir disable_stats =
          Vmm_lwt.connect Lwt_unix.PF_UNIX (Lwt_unix.ADDR_UNIX (socket_path s)) >>= fun x ->
          (match x with
          | Some x -> Lwt.return (Some x)
-         | None -> if (retries-1 != 0) then begin
+         | None -> if (retries != 0) then begin
```
hannesm, line 152:
```suggestion
         | None -> if retries <> 0 then begin
```

!= (and ==) is physical (in)equality, `<>` the normal inequality

daemon/albatrossd.ml
```diff
@@ -145,14 +145,39 @@ let jump _ influx tmpdir dbdir =
   | Error (`Msg msg) -> Logs.err (fun m -> m "bailing out: %s" msg)
   | Ok old_unikernels ->
     Lwt_main.run
-      (let unix_connect s =
-         Vmm_lwt.connect Lwt_unix.PF_UNIX (Lwt_unix.ADDR_UNIX (socket_path s))
+      (let rec unix_connect ~retries s =
+         Vmm_lwt.connect Lwt_unix.PF_UNIX (Lwt_unix.ADDR_UNIX (socket_path s)) >>= fun x ->
```
hannesm, line 149:
```suggestion
         Vmm_lwt.connect Lwt_unix.PF_UNIX (Lwt_unix.ADDR_UNIX (socket_path s)) >>= function
```

the `fun x -> match x with ..patterns..` can be condensed to `function ..patterns..`

daemon/albatrossd.ml
```diff
@@ -145,14 +145,39 @@ let jump _ influx tmpdir dbdir =
...
+         (match x with
+         | Some x -> Lwt.return (Some x)
+         | None -> if (retries != 0) then begin
+                     Logs.err (fun m -> m "unable to connect to %s, retrying in 5 seconds" (socket_path s));
+                     Lwt_unix.sleep 5.0 >>= fun () ->
+                     unix_connect ~retries:(retries-1) s
```
hannesm, line 155:
```suggestion
                     unix_connect ~retries:(retries - 1) s
```
I like spaces around operators :)

PR #40, opened 2020-11-10
Update Linux packaging

reynir:
This change updates the README and fixes some bugs.

I'm not happy with the systemd scripts:
* `albatross_log` doesn't create the log file if it's missing
* If `albatross_log` immediately crashes its dependencies (`albatross_daemon`) is still started, and then it crashes as well due to not being able to connect to the logger socket.


hannesm:
thanks.

PR #41, opened 2020-11-13
Ergonomics

reynir:
The commands `block`, `info`, and `policy` currently return an error in the case that there are no block devices, unikernels or policies respectively. At the client this results in a dangerous-looking `[WARNING]` when those commands are run on a fresh install.

This change makes the daemon return success with an empty list instead of an error. Instead, pretty printing is modified to make a note of the list being empty.

This PR contains two commits. The previous would make the daemon return a string with the message. I think it's cleaner if it's up to the client to handle empty lists. I can squash the commits into one if preferred.

### DIFF
--- a/src/vmm_commands.ml
+++ b/src/vmm_commands.ml
@@ -141,12 +141,17 @@ type success = [
 let pp_block ppf (id, size, active) =
   Fmt.pf ppf "block %a size %d MB active %B" Name.pp id size active
 
+let my_fmt_list empty pp_elt ppf xs =
+  match xs with
+  | [] -> Fmt.string ppf empty
+  | _ -> Fmt.(list ~sep:(unit "@.") pp_elt ppf xs)
+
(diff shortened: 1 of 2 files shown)


hannesm:
Thanks for your PR. Since the `Error` vs `Ok` leads to different exit codes of the client, I'm not entirely sure about the change in semantics. Maybe the commands are overloaded? Using `Block_info <id>` as example, a client may want to:
- list all block devices for their administrative domain (`Block_info []`) -- now if that's empty, an error is returned
- check before creating a specific block device (`my_block_device`) whether this already exists (`Block_info [ "my_block_device" ]`) -- now if this does not exist, an error is returned.

But considering that `Block_add` errors if the given block device already exists, and the above (execute block_info just before block_add) is racy, I agree with the proposed changes here (see minor nits inline). The remaining command, `Unikernel_get` , should be adapted to the same behaviour.

PR #43, opened 2020-11-18
Systemd socket activation

reynir:
This PR is just to keep track.

Things to be done:
- [x] Fix socket permissions
- [ ] Sort out dependencies. The individual services depend on the other services sockets instead of their services. I'm not sure it should be done this way. It seems to Work For Me™ so far though.
- [x] Pure OCaml implementation of `sd_listen_fds` so we don't depend on the Linux-specific `ocaml-systemd`. Or some other way around this issue.


packaging/Linux/albatross_log.service
```diff
@@ -4,17 +4,18 @@
 # systemctl edit albatross_log.service
 Description=Albatross log daemon (albatross_log)
 After=syslog.target albatross_console.service
-Requires=albatross_console.service
+Requires=albatross_console.socket albatross_log.socket
```
hannesm, line 7:
```suggestion
Requires=albatross_log.socket
```

the albatross_log service does not depend on the albatross_console service

src/vmm_lwt.mli
```diff
@@ -2,7 +2,7 @@
 
 val pp_sockaddr : Format.formatter -> Lwt_unix.sockaddr -> unit
 
-val server_socket : Vmm_core.service -> Lwt_unix.file_descr Lwt.t
+val server_socket : bool -> Vmm_core.service -> Lwt_unix.file_descr Lwt.t
```
hannesm, line 5:
```suggestion
val server_socket : systemd:bool -> Vmm_core.service -> Lwt_unix.file_descr Lwt.t
```

src/vmm_lwt.ml
```diff
@@ -12,20 +12,25 @@ let safe_close fd =
...
-  let s = Lwt_unix.(socket PF_UNIX SOCK_STREAM 0) in
-  Lwt_unix.set_close_on_exec s ;
-  let old_umask = Unix.umask 0 in
-  let _ = Unix.umask (old_umask land 0o707) in
-  Lwt_unix.(bind s (ADDR_UNIX name)) >|= fun () ->
-  Logs.app (fun m -> m "listening on %s" name);
-  let _ = Unix.umask old_umask in
-  Lwt_unix.listen s 1 ;
-  s
+let server_socket systemd sock =
```
hannesm, line 15:
```suggestion
let server_socket ~systemd sock =
```

hannesm:
this mostly looks fine - thanks for doing it.

I'm wondering that now the information about which socket name is used is in systemd (and the services are started accordingly). This also means that albatross_daemon that requires albatross_console.socket should take the socket name as argument, or not?

With this PR, the albatross_daemon will still use the hardcoded path from `tmpdir/util/console.sock` (as written in `socket_path` function in src/Vmm_core) -- does this matter (certainly not for a standard installation, but for someone who wants to modify the socket paths that makes a difference since they may think that they'd only need to adapt the systemd scripts)? Maybe a sentence in the Linux installation instructions is sufficient to not worry about that potential issue.

hannesm:
thanks a lot, merged

PR #45, opened 2020-11-25
Linux /proc/ stats

reynir:


stats/albatross_stats_pure.ml
```diff
@@ -99,9 +99,95 @@ let try_open_vmmapi pid_nic =
...
+    e (Int64.of_string_opt v)
+  in
+  let i64 s = try Ok (Int64.of_string s) with
+      Failure _ -> Error (`Msg "couldn't parse integer")
+  in
+  if List.length stat_vals >= 52 && List.length statm_vals >= 7 then
+    i64 (List.nth stat_vals 9) >>= fun minflt ->
+    i64 (List.nth stat_vals 11) >>= fun majflt ->
+    i64 (List.nth stat_vals 13) >>= fun utime -> (* divide by sysconf(_SC_CLK_TCK) *)
+    i64 (List.nth stat_vals 14) >>= fun stime -> (* divide by sysconf(_SC_CLK_TCK) *)
```
hannesm, line 161: Indeed we should use `_SC_CLK_TCK` -- below the `int64 * int` pairs for utime & stime are `timeval` structures -- i.e. represent seconds and microseconds. Maybe we should just revise it to a single int64 carrying the microseconds.

stats/albatross_stats_pure.ml
```diff
@@ -99,9 +99,95 @@ let try_open_vmmapi pid_nic =
...
+    i64 (List.nth stat_vals 35) >>= fun nswap -> (* not maintained, 0 *)
+    i64 (List.nth statm_vals 3) >>= fun tsize ->
+    i64 (List.nth statm_vals 5) >>= fun dsize -> (* data + stack *)
+    i64 (List.nth statm_vals 5) >>= fun ssize -> (* data + stack *)
+    assoc_i64 "voluntary_ctxt_switches" >>= fun nvcsw ->
+    assoc_i64 "nonvoluntary_ctxt_switches" >>= fun nivcsw ->
+    let rusage = { Stats.utime = (utime, 0) ; stime = (stime, 0) ; maxrss = rss ; ixrss = 0L ;
+         idrss = 0L ; isrss = 0L ; minflt ; majflt ; nswap ; inblock = 0L ; outblock = 0L ;
+         msgsnd = 0L ; msgrcv = 0L ; nsignals = 0L ; nvcsw ; nivcsw }
+    and kmem = { Stats.vsize; rss; tsize; dsize; ssize; runtime = 0L; cow = 0; start }
```
hannesm, line 173: in albatross_stat_client, I use the `runtime` (in FreeBSD this is "real time in microsec" (from struct kinfo_proc in sys/user.h)) -- it would be nice if Linux has a similar value, otherwise I guess we can add utime and stime for the total runtime.

hannesm:
looks good apart from the time resolution nits above.

stats/albatross_stats_pure.ml
reynir: Hmm, is `runtime` not real time since the process was started? If so, then adding `stime` and `utime` is not sound if the process is idle, I think.
I wrote down some notes with ideas how to get the value otherwise.

reynir:
I pushed a commit dividing stime and utime. I'm not too confident about the microseconds calculation. I'll have to revisit that tomorrow after I've had some sleep.

stats/albatross_stats_pure.ml
hannesm: no, it is the "CPU time used by this process in microseconds", I use the following to retrieve the CPU load (in percent): `(current_runtime - old_runtime) / interval [also in microseconds]` --> since we're single-core anyways, this gives me a percentage if applied on every interval.

if there's no interval, we can as well compute `runtime / (now () - start_time) [again in microseconds]` to compute the overall CPU usage since start in percent. does this make sense?

(I got the idea from the implementation of `ps` on BSD ;) -- the code above is in albatross_stat_client - I'll eventually play around to see whether using stime + utime shows the same numbers (or look into `ps` for Linux - how does this compute the `CPU %` column?))

hannesm:
Thanks! :) I'll re-review tomorrow and look into numbers.

stats/albatross_stats_pure.ml
hannesm: from procps source code https://github.com/warmchang/procps/blob/master/ps/output.c
```C
static int pr_cp(char *restrict const outbuf, const proc_t *restrict const pp){
  unsigned long long total_time;   /* jiffies used by this process */
  unsigned pcpu = 0;               /* scaled %cpu, 999 means 99.9% */
  unsigned long long seconds;      /* seconds of process life */
  total_time = pp->utime + pp->stime;
  if(include_dead_children) total_time += (pp->cutime + pp->cstime);
  seconds = cook_etime(pp);
  if(seconds) pcpu = (total_time * 1000ULL / Hertz) / seconds;
  if (pcpu > 999U) pcpu = 999U;
  return snprintf(outbuf, COLWID, "%3u", pcpu);
}
```
-> just use stime + utime :)

hannesm:
thanks, merged.

PR #47, opened 2020-11-26
Verify devices with manifest

### LINKED ISSUE #42: Check manifest for devices
Using `solo5-elftool query-manifest` we can check if the user has passed exactly all the devices required by the unikernel when using the `create` command, and report an error if they don't match.

reynir:
**WIP**

This change makes albatrossd check the manifest returned by `solo5-elftool query-manifest <image>` with the devices specified by the client and returns an error message if they don't match up.

I am thinking it may be worthwhile to check on the client end as well before the unikernel image is sent across the wire.

Fixes #42.

### DIFF
--- a/command-line/albatross_cli.ml
+++ b/command-line/albatross_cli.ml
@@ -102,7 +102,9 @@ let setup_log style_renderer level =
 
 let create_vm force image cpuid memory argv block_devices bridges compression restart_on_fail exit_codes =
   let open Rresult.R.Infix in
-  Bos.OS.File.read (Fpath.v image) >>| fun image ->
+  let img_file = Fpath.v image in
+  Bos.OS.File.read img_file >>= fun image ->
+  Vmm_unix.manifest_devices_match ~bridges ~block_devices img_file >>| fun () ->
   let image, compressed = match compression with
     | 0 -> Cstruct.of_string image, false
     | level ->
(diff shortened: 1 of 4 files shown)


hannesm:
Thanks for working on this. I think it makes a lot of sense to to this check in albatross_client as well.

On the server side, I wonder whether we want to ensure that the bridge interfaces are configured on the host system, and fail with an appropriate error message if not (AFAICT atm it just calls out to `ifconfig <bridge> addm <tap>` or `ip link set dev <tap> master <bridge>`).

Block devices are slightly different: albatross_daemon reads the `db_dir / block_sub` at startup and pretends only albatross_daemon will do any changes to that directory hierarchy (which I think is fine for now -- esp. since the block device part is not very much tested, so let's leave it in that state).

hannesm:
I implemented the suggestions above myself, and pushed two commits here.

reynir:
Looks good to me! Thanks.

reynir:
https://github.com/reynir/albatross/pull/1 makes it so `solo5-elftool` is not a hard dependency for the clients so clients can create unikernels without solo5-elftool.

hannesm:
Thanks @reynir -- I'm not sure whether this is the way to go -- since it results in albatross_client_* having different semantics (in terms of which errors are displayed / whether a create is successful) depending on what things are installed on the system. I find this irritating for a user, and am not sure whether it is worth the code complexity.

Issue #37, opened 2020-07-29
albatrossd did not clear up tap devices on shutdown


hannesm:
a `service albatross_daemon stop` killed all the unikernels, but left the tap devices in place. this was with 2d26a56c0d21bf9bf195a9dd448cf90f83ea2202. need to investigate (stress test) whether this is still an issue with more current albatross.

hannesm:
system log:
```
Jul 29 12:03:59 <kern.info> beast kernel: tap6: link state changed to DOWN
Jul 29 12:03:59 <daemon.notice> beast daemon[84677]: albatrossd: [WARNING] unikernel [vm: YYY] solo5 exit failure (1)
Jul 29 12:03:59 <daemon.notice> beast daemon[84638]: albatross_stats: [INFO] removing vmid [vm: ZZZ]
Jul 29 12:03:59 <daemon.notice> beast daemon[84638]: albatross_stats: [INFO] removing pid AAAA
```

the "solo5 exit failure (1)" is from Vmm_core.should_restart, but for some reason the waitpid handler was not executed (and the new albatross_log failed to read entries with version 0x2 /o\ -- unclear whether the Unikernel_stop log entry was dumped)

hannesm:
from the albatross log, there's only a few "stopped unikernel YYY" -- the `sleep 1` in https://github.com/hannesm/albatross/blob/1b1164166b409ec4daf6ea87d067031a7bc9973a/daemon/albatrossd.ml#L176-L186 should be revised (with appropriate waiters / inversion of control / eventually `Lwt.join`)

PR #50, opened 2020-12-01
Implement killall waiting

reynir:


daemon/albatrossd.ml
```diff
@@ -175,13 +175,11 @@ let jump _ systemd influx tmpdir dbdir retries enable_stats =
...
          Lwt_mutex.with_lock self_destruct_mutex (fun () ->
-             (if Vmm_vmmd.killall !state then
-                (* not too happy about the sleep here, but cleaning up resources
-                   is really important (fifos, vm images, tap devices) - which
-                   is done asynchronously (in the task waitpid() on the pid) *)
-                Lwt_unix.sleep 1.
-              else
-                Lwt.return_unit) >>= fun () ->
+             (let state', tasks = Vmm_vmmd.killall !state Lwt.task in
+              state := state';
```
reynir, line 179: I believe we should acquire the `create_lock` here so the waitpids know about the new waiters

hannesm:
this should fix #37 (or am I misguided?)

reynir:
Yes, this should ensure `Vmm_vmmd.handle_shutdown` is called which calls `Vmm_unix.free_system_resources`.

reynir:
One concern I have is `Vmm_vmmd.register` will overwrite any existing waiter. It's unclear to me if registering in `Vmm_vmmd.killall` may overwrite something important, or if there's a situation where something could overwrite our waiter after the `Vmm_vmmd.killall` call.

reynir:
I did a test on my machine creating 92 unikernels and then shutting down albatrossd using this branch. It took over 18 seconds to shut down. After the shutdown there were no `vmmtap`s lingering.
```shell
$ time sudo systemctl stop albatross_daemon.service 

real	0m18.202s
user	0m0.003s
sys	0m0.060s
```
I think it's reasonable to say it fixes #37.

(my router was not too happy getting 90 dhcp clients all of a sudden)

hannesm:
> One concern I have is `Vmm_vmmd.register` will overwrite any existing waiter. It's unclear to me if registering in `Vmm_vmmd.killall` may overwrite something important,

The only waiter to be registered is the `restart-on-failure` one. So indeed before destroying all unikernels we should overwrite all waiters :) (otherwise the restart-on-failure task may jump in and restart a unikernel).

> or if there's a situation where something could overwrite our waiter after the `Vmm_vmmd.killall` call.

This should not happen, as the only thing registering is in albatrossd ("if Unikernel.restart_handler config then Vmm_vmmd.register_restart ...") just after a unikernel was successfully started.

daemon/albatrossd.ml
```diff
@@ -175,13 +175,13 @@ let jump _ systemd influx tmpdir dbdir retries enable_stats =
...
-                Lwt_unix.sleep 1.
-              else
-                Lwt.return_unit) >>= fun () ->
+             Lwt_mutex.with_lock create_lock (fun () ->
+                 let state', tasks = Vmm_vmmd.killall !state Lwt.task in
+                 state := state';
+                 Lwt.return tasks) >>= fun tasks ->
+             Lwt_list.iter_s (fun exit_code ->
+                 exit_code >>= fun (_ : process_exit) -> Lwt.return_unit)
+               tasks >>= fun () ->
```
hannesm, line 184: is the `Lwt_list.iter_s` in this case a `Lwt.join`?

src/vmm_vmmd.mli
```diff
@@ -32,7 +32,7 @@ val handle_command : 'a t -> Vmm_commands.wire ->
    | `Wait_and_create of Name.t * (Name.t * Unikernel.config) ],
    Vmm_commands.res) result
 
-val killall : 'a t -> bool
+val killall : 'a t -> (unit -> 'b * 'a) -> 'a t * ('b list)
```
hannesm, line 35:
```suggestion
val killall : 'a t -> (unit -> 'b * 'a) -> 'a t * 'b list
```

daemon/albatrossd.ml
reynir: It is *almost*. `Lwt.join : unit Lwt.t list -> unit Lwt.t` while the above has type `process_exit Lwt.t list -> unit Lwt.t`. I could rewrite it, I believe, to:
```OCaml
Lwt.join (List.map (Lwt.map ignore) tasks)
```

hannesm:
thanks, I pushed a commit to first shut down the server socket, and then to destroy the unikernels.... I think this is good to go (let's wait for CI though) -- and please comment if you see any problems with the revised shutdown order..

reynir:
I think it makes more sense to close the socket first, yes.

PR #57, opened 2020-12-02
use the conf-libnl3 opam package

hannesm:


albatross.opam
```diff
@@ -11,6 +11,8 @@ depends: [
   "dune"
   "dune-configurator"
   "conf-pkg-config" {build}
+  "conf-libnl3" {os = "linux"}
+  "conf-linux-libc-dev" {os = "linux"}
```
kit-ty-kate, line 15: I haven't tried but I think https://github.com/roburio/albatross/pull/58 might be enough to not require `conf-linux-libc-dev`.

hannesm: thanks @kit-ty-kate

PR #58, opened 2020-12-02
Avoid using a deprecated include on Linux

kit-ty-kate:
From `/usr/include/sys/sysctl.h` on my Debian Sid:
```
#warning "The <sys/sysctl.h> header is deprecated and will be removed."
```


reynir:
Thanks! I realized we actually don't use sysctl.h in Linux, so I moved it down to `#ifdef __FreeBSD__`.

PR #64, opened 2021-01-04
FreeBSD albatross_daemon rc script: extend PATH, revise unikernel_get and unikernel_info commands

### LINKED ISSUE #49: improve block device support
sync them with network devices. so far, the policy contains "block device capacity" and "bridge names" that are allowed for the arc.

now, a unikernel_create gets a list of <service>:<bridge option>, i.e. "--bridge external:public" or "--bridge internal:management" (or "--bridge public") -- meaning in order: acquire a tap device on bridge "public", and pass "--net:external=<tap>" to the tender; acquire a tap device on bridge "management" and pass "--net:internal=<tap>" to the tender; acquire a tap device on bridge "public" and pass "--net:public=<tap>" to the tender.

the block devices are all kept in "dbdir/block/", so block_add & remove take the arc (from ca chain / header) and create the specified device. this directory is read at startup. now, if a unikernel image 'test' requires "block device 'foo'" (in the com.example domain): dbdir/block/com.example.foo is used -- and passed to the tender -- there's no way to use block device "dbdir/block/com.example.bar" instead. and this is what should be cleaned up (similar to how bridges are passed by having a pair with a string option: "--block foo:bar" should be possible)

### LINKED ISSUE #61: feature request: retrieve the SHA256 of a running unikernel
at the moment, there is `info` for a unikernel which returns the unikernel configuration without the unikernel image itself. and there is `get` that returns the entire configuration (including image).

in the light of reproducible builds etc. it'd be great to get the uniquely identified sha-256 hash of the running (and uncompressed) binary of a unikernel. this would ease inventory management. the hash could be computed upon unikernel startup (from the uncompressed data dumped to disk) to avoid re-computing the hash on every request.

I'd appreciate it being part of the `info` subcommand (meaning revising the Unikernel.config record and ASN grammar -- but note that it should not be provided by albatross_client_*, but instead computed by the server! so eventually the `get` and `info` subcommands should use some other data structure than the create/force-create subcommands)

### LINKED ISSUE #63: locations of solo5 binaries
I got some (spurious?) errors when solo5-elftool and solo5-hvt are installed by the system in PATH (instead of dbdir). check that the code does not require them to be in dbdir.

hannesm:
This now includes /usr/local/bin and /usr/local/sbin in the PATH variable. The
solo5 binaries will likely be installed in these locations, and albatross_daemon
needs to be able to find them. Fixes #63.

The second commit revises unikernel_info and unikernel_get commands and fixes #61 in a backwards-compatible way.

The third commit revises the block device handling and fixes #49 in a backwards-compatible way.

### DIFF
--- a/src/vmm_asn.ml
+++ b/src/vmm_asn.ml
@@ -295,14 +295,14 @@ let log_event =
         | Some block -> [ block, Name.block_name name block ]
       and taps = List.map (fun tap -> tap, tap) taps
       in
-      `Unikernel_start (name, pid, taps, blocks)
+      `Unikernel_start (name, Cstruct.empty, pid, taps, blocks)
     | `C2 `C1 (name, pid, taps, blocks) ->
       let blocks = List.map (fun (name, dev) ->
           name, match Name.of_string dev with
           | Error `Msg msg -> Asn.S.error (`Parse msg)
           | Ok id -> id) blocks
       in
-      `Unikernel_start (name, pid, taps, blocks)
+      `Unikernel_start (name, Cstruct.empty, pid, taps, blocks)
     | `C1 `C5 (name, pid, status) ->
       let status' = match status with
         | `C1 n -> `Exit n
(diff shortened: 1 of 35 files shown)


src/vmm_asn.ml
```diff
@@ -508,7 +538,11 @@ let unikernel_config =
          @ (required ~label:"fail-behaviour" fail_behaviour)
          @ (required ~label:"cpuid" int)
          @ (required ~label:"memory" int)
-         @ (optional ~label:"blocks" (my_explicit 0 (set_of utf8_string)))
+         @ (optional ~label:"blocks"
+              (my_explicit 0 (set_of
```
reynir, line 554: Here we use `set_of` while for bridges we use `sequence_of` for essentially the same structure. Semantically, I think this makes sense. Maybe we should change bridges to use `set_of` as well if we're making changes anyway?

reynir (commented):
This looks good to me!

src/vmm_asn.ml
hannesm: well spotted, indeed this is a relict from back when solo5 didn't had --net:<name>=<tap>, and the order was important. I changed it here and in unikernel_info as well, and force-pushed.

PR #65, opened 2021-01-05
Add ASN test

reynir:
The test is just enough to trigger exception Asn_core.Ambiguous_syntax.


hannesm:
nice, then we may want to adapt albatross.opam to run the test (by adding ` ["dune" "runtest" "-p" name "-j" jobs] {with-test}` to the `build:` stanza) :)

hannesm:
thanks

PR #67, opened 2021-01-06
improve block handling:

hannesm:
- Vmm_resources.check_vm needs to check for the right block device
- the solo5-hvt/spt argument is --block:<>=<>, not --disk
- Vmm_tls: allow block commands
- Vmm_unix.find_block_devices creates dbdir / block_sub if it does not exist
  (previously, this lead to a crash of albatrossd)

In addition, Vmm_resources.check_policy returns more detailed error messages.

### DIFF
--- a/src/vmm_unix.ml
+++ b/src/vmm_unix.ml
@@ -140,9 +140,12 @@ let dump, restore =
 
 let block_sub = "block"
 
+let block_dir () =
+  Fpath.(!dbdir / block_sub)
+
 let block_file name =
   let file = Name.to_string name in
-  Fpath.(!dbdir / block_sub / file)
+  Fpath.(block_dir () / file)
 
 let rec mkfifo name =
   try Unix.mkfifo (Fpath.to_string name) 0o640 with
@@ -310,7 +313,10 @@ let exec name (config : Unikernel.config) bridge_taps blocks digest =
            "--net:" ^ bridge ^ "=" ^ tap,
            "--net-mac:" ^ bridge ^ "=" ^ Macaddr.to_string mac)
           bridge_taps)
-  and blocks = List.map (fun (name, dev) -> "--disk:" ^ name ^ "=" ^ Fpath.to_string (block_file dev)) blocks
+  and blocks =
+    List.map (fun (name, dev) ->
+        "--block:" ^ name ^ "=" ^ Fpath.to_string (block_file dev))
+      blocks
   and argv = match config.Unikernel.argv with None -> [] | Some xs -> xs
   and mem = "--mem=" ^ string_of_int config.Unikernel.memory
   in
@@ -358,6 +364,10 @@ let create_block name size =
   Bos.OS.File.exists block_name >>= function
   | true -> Error (`Msg "file already exists")
   | false ->
+    let dir = block_dir () in
+    (Bos.OS.Path.exists dir >>= function
+      | false -> Bos.OS.Dir.create ~mode:0o700 dir
+      | true -> Ok true) >>= fun _ ->
     let fd = Unix.(openfile (Fpath.to_string block_name) [O_CREAT] 0o600) in
     close_no_err fd ;
     bytes_of_mb size >>= fun size' ->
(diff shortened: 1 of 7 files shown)


src/vmm_unix.ml
```diff
@@ -374,23 +377,26 @@ let mb_of_bytes size =
...
-        Bos.OS.Path.stat path >>= fun stats ->
-        match mb_of_bytes stats.Unix.st_size, Name.of_string (Fpath.to_string file) with
-        | Error (`Msg msg), _ ->
-          Logs.warn (fun m -> m "file %a size error: %s" Fpath.pp path msg) ;
-          Ok acc
-        | _, Error (`Msg msg) ->
-          Logs.warn (fun m -> m "file %a name error: %s" Fpath.pp path msg) ;
+  Bos.OS.Path.exists blockdir >>= function
+  | false -> Bos.OS.Dir.create blockdir >>| fun _ -> []
+  | true ->
```
reynir, line 382: Maybe replace with:

``` OCaml
(Bos.OS.Path.exists blockdir >>= function
  | false -> Bos.OS.Dir.create blockdir >>| fun _ -> ()
  | true -> Ok ()) >>= fun () ->
```

It adds an unnecessary system call, but the diff is smaller and we catch the (probably highly unlikely) case where someone creates a file inbetween the exists call and create. I don't have strong feelings about this...

src/vmm_unix.ml
```diff
@@ -374,23 +377,26 @@ let mb_of_bytes size =
...
-      | true ->
-        Bos.OS.Path.stat path >>= fun stats ->
-        match mb_of_bytes stats.Unix.st_size, Name.of_string (Fpath.to_string file) with
-        | Error (`Msg msg), _ ->
-          Logs.warn (fun m -> m "file %a size error: %s" Fpath.pp path msg) ;
-          Ok acc
-        | _, Error (`Msg msg) ->
-          Logs.warn (fun m -> m "file %a name error: %s" Fpath.pp path msg) ;
+  Bos.OS.Path.exists blockdir >>= function
+  | false -> Bos.OS.Dir.create blockdir >>| fun _ -> []
```
reynir, line 381: We probably want to pass `~mode:0o700` or something:
`[create ~path ~mode dir] creates, if needed, the directory [dir] with file permission [mode] (defaults [0o755] readable and traversable by everyone, writeable by the user).`

reynir (changes requested):
The permissions for the directory is maybe a bit too permissive.

The Linux packaging script could be updated to not create the block directory (I think that's how the FreeBSD packaging works?)

src/vmm_unix.ml
hannesm: good point. I guess for safety the case where someone removes the block subdirectory while albatrossd is executing also needs to be handled (when `open` returns ENOENT)

src/vmm_unix.ml
hannesm: albatrossd (executing this function) is running as root (for now, in order to create virtual machines). I'm not entirely sure about the permissions that should be in place for block devices (and unikernel image files in /var/run/albatross)... there's no need for others to read/write/execute any of that...

reynir: I would expect the block devices to not be readable by any other users, and probably the same for the image files.

Should we make a separate issue with permissions review and ignore it for now?

hannesm: maybe we leave this PR open until all the permission and exceptions are dealt with, shouldn't take too long, and better be done with it.

hannesm:
> The Linux packaging script could be updated to not create the block directory

right, there's the question what the packaging scripts should do, and what albatrossd should do. I'm also fine with relying on packaging scripts to create the necessary directories -- but nevertheless albatrossd should check their existence and complain (instead of fail via an exception) if the on-disk layout / permissions are not suitable for its operations.

hannesm:
I revised this PR, and think it is good to be merged now. The FreeBSD installation script now creates the block directory, and any invocation of "create_block" ensures the block directory to be present (otherwise, it is created with permissions 700).

src/vmm_unix.ml
hannesm: now, the create_block creates the block directory if it does not exist -- and uses permissions 0700

src/vmm_unix.ml
hannesm: this is now solved - installation scripts create (with permission 0700), the create_block creates with 0700 (files with 0600).

PR #68, opened 2021-01-13
use the last version of decompress

dinosaure:
Due to the release of `git.3.0.0` which includes the new version of `decompress`, it could be interesting to upgrade `albatross` to the last vesion of `decompress` (`>= 1.0.0`).

### DIFF
--- a/albatross.opam
+++ b/albatross.opam
@@ -30,7 +30,8 @@ depends: [
   "mirage-crypto-rng" {>= "0.8.0"}
   "asn1-combinators" {>= "0.2.0"}
   "duration"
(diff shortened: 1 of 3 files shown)


hannesm:
thanks. just one question: is the compression done by decompress < 1.0.0 compatible with >= 1.2.0?

dinosaure:
> thanks. just one question: is the compression done by decompress < 1.0.0 compatible with >= 1.2.0?

Yes, the format does not change, it's only implementation and speed improvement (in details, we still respect RFC1951 + zlib header).

hannesm:
thanks, pushed as 38f5085 with the tab character in src/dune converted to spaces.

PR #73, opened 2021-05-09
packaging linux: albatross_stat -> albatross_stats

smorimoto:
As the title says.


hannesm:
thanks, squashed and merged.

PR #79, opened 2021-06-19
Add albatross user and allow user into the group to use albatross

dinosaure:
It's a small improvement but when we want to install albatross, my configuration is like:
- I have an user `mirage`
- The script should create the `albatross` user
- `mirage` can be a part of the `albatross` group
- `mirage` should be allowed to use _sockets_

This patch allows the `albatross` group to use _sockets_ and create the required `albatross` user for the installation. In the opposite, the daemon fall back as a `root` user and _sockets_ got permissions only for the `root` user.


hannesm:
Thanks, I merged this manually and adapted the `useradd` (by specifying home, login shell, and also create a group).

PR #81, opened 2021-09-03
Allow any from the albatross group to manipulate the vmmd.sock

dinosaure:
Close #79, it permits to use `albatross-client-local` with an user which is not `root` but it is a part of the `albatross` group.


hannesm:
thanks

PR #82, opened 2021-09-03
Add a way to create a block device from a file

dinosaure:
Currently, it's not possible to create a new block device from a file - and then, use it for many unikernels. This is a fast patch, I don't really if it's the right way to implement it, but, at least, it compiles.

### DIFF
--- a/client/albatross_client_bistro.ml
+++ b/client/albatross_client_bistro.ml
@@ -138,8 +138,8 @@ let stats _ endp cert key ca name =
 let block_info _ endp cert key ca block_name =
   jump endp cert key ca block_name (`Block_cmd `Block_info)
 
-let block_create _ endp cert key ca block_name block_size =
-  jump endp cert key ca block_name (`Block_cmd (`Block_add block_size))
+let block_create _ endp cert key ca block_name block_src block_size =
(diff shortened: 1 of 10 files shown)


dinosaure:
Currently, this patch seems to work when the saved block device has the same hash than the source file.

hannesm:
Since this changes the wire format (Block.add used to be an int, now a sequence of optional string, required int), this needs to be done more carefully... but the overall idea is fine (maybe a block set / block get command should be implemented in addition to the existing (previous) add).

dinosaure:
A question comes with this PR. Is it possible to use the same block for multiple unikernels (in read-only mode)?

hannesm:
> Is it possible to use the same block for multiple unikernels (in read-only mode)?

no, the block devices, as defined by solo5 and mirageos are read-write. only one unikernel may attach the very same block device.

hannesm:
another issue is that commands are transported via X.509 client certificates that are limited in size (< 16MB). So a Block set command could only work for small block devices. Maybe it is better to develop a unikernel that retrieves or generates data from somewhere and dumps it on a block device. Though, when the block device is used for small (configuration) information, 16 MB is sufficient.

hannesm:
thanks for your work. please have a look at #85 for an updated version.

PR #85, opened 2021-09-10
implement block_set and block_dump

hannesm:
also, revise block_add to optionally take data. alternative to #82

//cc @dinosaure would this suit your needs?


hannesm:
this extends the wire protocol with some backwards compatibility:
- a new server will understand the old client (`` `C2 of int ``  is Block_add (size, None))
- a new client will issue `` `C4 `` for Block_add, which an old server will reject (but here, the client can be downgraded)

Issue #93, opened 2022-01-25
use solo5-elftool opam package instead of calling out to solo5-elftool


hannesm:
this removes the dependency in albatross-client-bistro -- it does not make a difference for the albatross daemon (which requires solo5-hvt/solo5-spt)

//cc @reynir

hannesm:
fixed by #94

PR #94, opened 2022-01-25
Use Solo5_elftool instead of calling binary

reynir:
Uses Obj.magic to convert between the underlying bigarray in Cstruct and Owee_buf.t.

This is work I made before Christmas and quickly brushed up to make it compile, and added some more type ascriptions around Obj.magic.

Addresses #93


hannesm:
thanks a lot!

PR #95, opened 2022-01-27
debian: install daemons to /usr/libexec/albatross/ to avoid accidental execution

### LINKED ISSUE #91: Avoid daemons to override sockets of already executing daemons
To avoid accidental executions of albatross_console etc. (which will create and claim the socket). Eventually these programs should as well not start up if the socket is already there and opened (but that may lead to rather intricate semantics).

hannesm:
fixes #91


hannesm:
this removes the `install.sh` from the packging/Linux directory, since its usefulness is questionable - and I prefer to build system packages and maintain paths etc. only once.

packaging/debian/create_package.sh
```diff
@@ -9,16 +9,16 @@ fi
 
 basedir=$(realpath "$(dirname "$0")"/../..)
 bdir=$basedir/_build/install/default/bin
-tmpd=$basedir/_build/stage
+gtmpd=$basedir/_build/stage
```
reynir, line 12: Looks like this was a typo not meant to be committed.

reynir (changes requested):
I will fix the typo now. It looks good to me otherwise. Thanks!

hannesm:
thanks, indeed a typo and me not looking through the diff completely.

reynir (approved):

PR #103, opened 2022-03-01
Linux: add albatross_influx systemd service script

hannesm:


packaging/debian/create_package.sh
```diff
@@ -47,6 +47,8 @@ do
     install -m 0644 $basedir/packaging/Linux/$f.service $systemddir/$f.service;
     install -m 0644 $basedir/packaging/Linux/$f.socket $systemddir/$f.socket;
 done
+install -m 0644 $basedir/packaging/Linux/albatross_influx.service \
```
reynir, line 50: Thinking aloud: We don't need `albatross_influx.socket` because albatross-influx does not listen for connections. Instead, it connects to the stats daemon and forwards the stats to influx.

packaging/Linux/albatross_influx.service
```diff
@@ -0,0 +1,15 @@
...
+# to create an override configuration:
+# systemctl edit albatross_influx.service
+Description=Albatross influx reporting daemon (albatross-influx)
+Requires=albatross_stats.socket
+After=syslog.target
+
+[Service]
+Type=simple
+User=albatross
+ExecStart=/usr/libexec/albatross/albatross-influx --influx=127.0.0.1 --tmpdir="%t/albatross/" -vv
```
reynir, line 12: We should at some point revise `-vv`, but we have it already in the other service files.

reynir (approved):

PR #107, opened 2022-03-25
Custom mac address

reynir:
This PR extends the `--net` command line argument to take an optional mac address separated by an ampersand. The format is `--net=service:bridge@00:11:22:33:44:55`.

Things to do:
- [x] Fix pretty printing of optional mac address.
- [x] Fix the ASN.1. I'm confused how much backwards compatibility we need.
- [x] Actually pass `--net-mac` to solo5.
- [ ] ~~Consider how it fits in with the resource policies. Likely, we want to be able to put restrictions on the use of custom mac addresses.~~

### DIFF
--- a/command-line/albatross_cli.ml
+++ b/command-line/albatross_cli.ml
@@ -320,7 +320,8 @@ let args =
   Arg.(value & opt_all string [] & info [ "arg" ] ~doc)
 
 let colon_separated_c =
-  let parse s = match String.split_on_char ':' s with
+  let parse s =
+    match String.split_on_char ':' s with
     | [ a ; b ] -> Ok (a, Some b)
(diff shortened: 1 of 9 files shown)


reynir:
It occurred to me that we would want to be able to impose a policy on custom mac addresses. Especially in a multi-tenant setup. As a first iteration a binary 'allow all/disallow custom mac' is probably fine. Added a sub task for it.

hannesm:
May be sensible to add a test with a unikernel that has a mac address specified.

hannesm:
about resource policies: since tenants can be restricted to bridges, that's sufficient -- it is anyways the case that any unikernel can send out any ethernet frames... so nothing to worry about

PR #108, opened 2022-03-28
Fix 105

reynir:
I found that systemd would create `/run/albatross/fifo/` with permissions `drwxrws---  root root`. A new commit is added where Group is set to albatross in all systemd service files. The group albatross should be implied by `User=albatross` so it should only be strictly necessary to put it in albatross_daemon.service.


reynir:
Hmm, I'm not sure why the previous commit is in here...

reynir:
OK a rebase fixed it

hannesm:
thanks

PR #110, opened 2022-03-31
Vmm_resources: check_vm/check_block: locate policy (may be not parent name)

hannesm:
When deploying the unikernel foo.bar, policy in bar is looked for.
When deploying the unikernel a.foo.bar, policy in foo.bar is looked for.

But since unikernel names are <subCA>.<name>, and "name" may contain any ".",
it is crucial to check the policy of bar when deploying a.foo.bar.


hannesm:
This is a workaround, I'm more eager to develop a real fix for the confusion.

Issue #115, opened 2022-08-19
exception


hannesm:
somehow on FreeBSD, the albatross-stat ran into `Fatal error: exception Invalid_argument("List.combine")`, which in turn resulted in `albatrossd` unable to send more data to the stats daemon (and thus, a failure without cleanup of tap devices etc.).

reynir:
Should we reopen this? In ee58e29bc25cbb67fb0439b59d4a87ce0b66bcdb I referenced this commit because it might fix the issue, and github is eager to close issues mentioned in commits...

hannesm:
I think this is fine, since the List.combine is gone now :)

PR #116, opened 2022-08-22
Refactor albatross stats stubs

reynir:
In #115 an exception from `List.combine` is raised. With `git grep` I found two sites that call `List.combine`. One place it's called with `taps` and `vm.config.bridges`. This should be fine since the former is constructed from the latter.
The other place is in `albatross_stats_pure.ml` in order to associate a statistic / counter value with a string description. Previously, the statistics description strings was fetched once when opening the first VM. From what I could gather the number of VM statistic counters depends on the number of CPUs assigned, so my hypothesis is the number of counters for the first VM may have differed from a later VM.

This patch makes `vmmanage_vmmapi_stats` return `(string * int64) list`  instead of a `int64 list` where the string is the description fetched with `vm_get_stat_desc` at each `vmmanage_vmmapi_stats` call. This may be less efficient.

Ideally, we would somehow verify my hypothesis, and maybe benchmark the performance impact of this patch.

### DIFF
--- a/stats/albatross_stat_client.ml
+++ b/stats/albatross_stat_client.ml
@@ -26,8 +26,7 @@ let timer pid vmmapi interval =
   | None -> ()
   | Some vmctx -> match wrap vmmapi_stats vmctx with
     | None -> Logs.app (fun m -> m "no vmctx stats")
-    | Some st ->
-      let all = List.combine !descr st in
+    | Some all ->
       Logs.app (fun m -> m "bhyve stats %a" Stats.pp_vmm_mem all)
 
 let jump _ pid name interval tmpdir =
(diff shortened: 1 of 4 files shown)


hannesm:
Thanks, I thought maybe instead of `let descr = ref []` we could use `let descr = lazy (wrap vmmapi_statnames ctx)`, and apply `Lazy.force`. This way, ti would be evaluated only once, and the Lazy module would take care that it is good.

hannesm:
This looks fine, I pushed a commit that guard the statistics gathering of bhyve debug counters optional (via a flag passed to albatross-stats). WDYT?

reynir:
Nice, it looks good to me

PR #119, opened 2022-10-06
tls-endpoint: Listen on systemd socket

Julow:
Add support to `albatross-tls-endpoint` for listening on a systemd socket. This is useful to let systemd open priviledge ports safely.

The '--port' option is kept for compatibility and could also be added to the other commands thanks to a reusable cmdliner term.

### DIFF
--- a/command-line/albatross_cli.ml
+++ b/command-line/albatross_cli.ml
@@ -435,6 +435,25 @@ let systemd_socket_activation =
     let doc = "Pass this flag when systemd socket activation is being used" in
     Arg.(value & flag & info [ "systemd-socket-activation" ] ~doc)
 
+let port_or_socket ~default_port =
+  let open Term in
+  let port =
+    let doc = "TCP listen port." and absent = string_of_int default_port in
+    Arg.(value & opt (some int) None & info [ "port" ] ~doc ~absent)
+  in
+  term_result
+    (const (fun port socket ->
+         match (port, socket) with
+         | Some _, true ->
+             Error
+               (`Msg
+                 "Options --port and --systemd-socket-activation are not \
+                  compatible")
+         | None, true -> Ok `Systemd_socket
+         | Some p, false -> Ok (`Port p)
+         | None, false -> Ok (`Port default_port))
+    $ port $ systemd_socket_activation)
+
 let pub_key_type =
   let doc = "Asymmetric key type to use" in
   Arg.(value & opt (Arg.enum X509.Key_type.strings) `ED25519 & info [ "key-type" ] ~doc)
--- a/src/vmm_lwt.ml
+++ b/src/vmm_lwt.ml
@@ -12,25 +12,36 @@ let safe_close fd =
     (fun () -> Lwt_unix.close fd)
     (fun _ -> Lwt.return_unit)
 
-let server_socket ~systemd sock =
-  if systemd
-  then match Vmm_unix.sd_listen_fds () with
-    | Some [fd] -> Lwt.return (Lwt_unix.of_unix_file_descr fd)
-    | _ -> failwith "Systemd socket activation error" (* FIXME *)
-  else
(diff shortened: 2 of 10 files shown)


hannesm:
I think this is fine, but would like to hear from @reynir before merging.

src/vmm_lwt.ml
```diff
@@ -12,25 +12,36 @@ let safe_close fd =
...
+  setsockopt s SO_REUSEADDR true ;
+  setsockopt s IPV6_ONLY false ;
+  bind s (ADDR_INET (Unix.inet_addr_any, port)) >>= fun () ->
+  listen s 10 ;
+  Lwt.return s
+
+let systemd_socket () =
+  match Vmm_unix.sd_listen_fds () with
+  | Some [ fd ] -> Lwt.return (Lwt_unix.of_unix_file_descr fd)
+  | _ -> (* FIXME *) failwith "Systemd socket activation error"
```
reynir, line 28: Here we still expect exactly one systemd activation socket. I believe systemd supports passing more than one socket, but I am unsure how they are ordered. I am wondering if we can somehow ensure that if we in the future expect more than one socket that we don't accidentally reuse the same socket for two different things.

As stated, none of our services ever listen on more than one socket currently, so this is not a problem.

src/vmm_lwt.mli
```diff
@@ -2,7 +2,14 @@
...
 
-val server_socket : systemd:bool -> Vmm_core.service -> Lwt_unix.file_descr Lwt.t
+(** Listen on a port. *)
+val port_socket : int -> Lwt_unix.file_descr Lwt.t
+
+(** Listen on a socket passed by systemd through a variable. *)
+val systemd_socket : unit -> Lwt_unix.file_descr Lwt.t
+
+(** Listen on a socket bound to a service. *)
+val service_socket : Vmm_core.service -> Lwt_unix.file_descr Lwt.t
```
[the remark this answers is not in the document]
reynir: Yes, I think this is good. It's too much magic if we expect one function to do everything :-)

command-line/albatross_cli.ml
```diff
@@ -435,6 +435,25 @@ let systemd_socket_activation =
...
+         match (port, socket) with
+         | Some _, true ->
+             Error
+               (`Msg
+                 "Options --port and --systemd-socket-activation are not \
+                  compatible")
+         | None, true -> Ok `Systemd_socket
+         | Some p, false -> Ok (`Port p)
+         | None, false -> Ok (`Port default_port))
+    $ port $ systemd_socket_activation)
```
reynir, line 455: Nice.

reynir (approved):
I think this change looks good. I only have minor comments that are not blocking a merge.

What would be nice is to add `albatross_tls_endpoint.service` and `albatross_tls_endpoint.socket` examples.

src/vmm_lwt.ml
Julow: The [manual](https://www.freedesktop.org/software/systemd/man/sd_listen_fds.html) says:

> If multiple socket units activate the same service, the order of the file descriptors passed to its main process is undefined.

Quite crazy :p but if several sockets are defined in one unit file, the order is defined. There can only be one activation socket safely it seems (the manual recommend to check the type of the socket lol).

The `LISTEN_FDNAMES` variable contains the name of the sockets, colon-separated.

> It is recommended to rely on named file descriptors only if identification via sd_is_socket() and related calls is not sufficient.

Julow:
The systemd service for the TLS endpoint needs to pass the public/private keys as argument. I'm not sure how to package that? The nixos module is an example :)

reynir:
I added systemd examples for albatross-tls-endpoint. If you could take a look and see if it looks reasonable I'll merge this. Sorry for the long process.

Julow:
Thanks! No need to make an example that runs immediately indeed.

PR #123, opened 2022-10-24
The table of interface can be sparse

dinosaure:
Currently, we assume that the last ID of interfaces is equal to the number of interfaces available. But the documentation on FreeBSD and Linux say that the table of interface can be sparse. This patch is a naive one which try any ID until we found what we want or we discover all interfaces available.

The patch fix an issue on Linux (EMFILE) and properly close the `nl_socket` in any situation (even if the function fails).

### DIFF
--- a/stats/albatross_stats_stubs.c
+++ b/stats/albatross_stats_stubs.c
@@ -229,10 +229,14 @@ CAMLprim value vmmanage_sysctl_ifcount(value unit) {
   if (err < 0)
     uerror("nl_connect", Nothing);
   err = rtnl_link_alloc_cache(nl_sock, AF_UNSPEC, &link_cache);
-  if (err < 0)
+  if (err < 0) {
+    nl_close(nl_sock);
     uerror("rtnl_link_alloc_cache", Nothing);
+  }
+  long items = nl_cache_nitems(link_cache);
+  nl_close(nl_sock);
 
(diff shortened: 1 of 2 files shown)


stats/albatross_stats_stubs.c
```diff
@@ -229,10 +229,14 @@ CAMLprim value vmmanage_sysctl_ifcount(value unit) {
   if (err < 0)
     uerror("nl_connect", Nothing);
   err = rtnl_link_alloc_cache(nl_sock, AF_UNSPEC, &link_cache);
-  if (err < 0)
+  if (err < 0) {
+    nl_close(nl_sock);
     uerror("rtnl_link_alloc_cache", Nothing);
+  }
+  long items = nl_cache_nitems(link_cache);
```
reynir, line 236:
```suggestion
  long nitems = nl_cache_nitems(link_cache);
```

hannesm:
I'd prefer two options:
- in sysctl_ifcounr return the highest index that is available (as done on FreeBSD) - but it seems there's no easy path to that (apart from iterating over the entire cache)
- use a different C stub and call the rtnl_link_get_by_nane, and avoid the index dance (that seems to be not necessary on Linux anyways)

hannesm:
please take a look (and eventually try it out #124) :) thanks (also for closing the netlink sockets).

hannesm:
included and merged as #125

PR #125, opened 2022-10-24
Use get_ifindex_by_name

reynir:
This is an alternative to #123 and #124 that expects from the C stubs a function `get_ifindex_by_name`. This simplifies the OCaml code, and allows us to use `rtnl_link_get_by_name` on Linux, and on FreeBSD the ifindex dance is moved to the C stub.

### DIFF
--- a/stats/albatross_stats_stubs.c
+++ b/stats/albatross_stats_stubs.c
@@ -297,16 +297,19 @@ CAMLprim value vmmanage_sysctl_ifdata(value num) {
   if (nl_sock == 0)
     uerror("nl_socket_alloc", Nothing);
   err = nl_connect(nl_sock, NETLINK_ROUTE);
-  if (err < 0)
+  if (err < 0) {
+    nl_socket_free(nl_sock);
     uerror("nl_connect", Nothing);
+  }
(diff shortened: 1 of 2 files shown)


stats/albatross_stats_stubs.c
```diff
@@ -150,23 +150,47 @@ CAMLprim value vmmanage_vmmapi_stats (value octx) {
...
+  name_ifcount[2] = NETLINK_GENERIC;
+  name_ifcount[3] = IFMIB_SYSTEM;
+  name_ifcount[4] = IFMIB_IFCOUNT;
+  dlen = sizeof(ifcount);
+
+  if (sysctl(name, nitems(name_ifcount), &ifcount, &dlen, NULL, 0) != 0)
     uerror("sysctl", Nothing);
 
-  CAMLreturn(Val_long(data));
+  for (int idx = 1; idx <= ifcount; idx++) {
```
hannesm, line 178: maybe smarter to go from last index (ifcount) to 1 (or 0?)? The intuition is that this code is called for newly created virtual machines, that use fresh tap devices, and I expect the tap device numbering to be linear and monotonically increasing.

dinosaure (approved):

reynir:
\o/ I'm happy with the much simpler code we have now.

PR #126, opened 2022-10-25
Remove includes that are not necessary

reynir:
They were introduced in a commit in #125 but were not removed when the code was changed. It builds on Linux.


hannesm (approved):
good to merge if CI is green

Issue #118, opened 2022-10-05
add a restart command


hannesm:
to start a running unikernel from the beginning

PR #128, opened 2022-10-28
Minimal support for macOS

samoht:
Enough to run `albatross-client-local` without errors.


hannesm:
Thanks, I added GitHub action on macos so that the compilation is at least tested.

samoht:
When do you think you could make a new release with this fix? Thanks

hannesm:
> When do you think you could make a new release with this fix? Thanks

The missing piece is #134, but it may be worth to think whether #118 and/or #129 is important enough. Likely this week. Thanks for asking.

PR #133, opened 2022-11-05
update rc scripts / systemd files with the revised dependency

hannesm:
the dependencies are inverted now between albatross-daemon and albatross-stats

### DIFF
--- a/packaging/FreeBSD/MANIFEST
+++ b/packaging/FreeBSD/MANIFEST
@@ -1,5 +1,5 @@
 name:         albatross
-version:      %%VERSION_NUM%%
+version:      "%%VERSION_NUM%%"
 origin:	      local/albatross
 comment:      Albatross: orchestrate and manage MirageOS unikernels with Solo5
 www:	      https://github.com/roburio/albatross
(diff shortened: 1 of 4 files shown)


packaging/FreeBSD/MANIFEST
```diff
@@ -1,5 +1,5 @@
 name:         albatross
-version:      %%VERSION_NUM%%
+version:      "%%VERSION_NUM%%"
```
hannesm, line 2: This avoids the (cirrus) CI issue `pkg: manifest parsing error: error while parsing <unknown>: line: 2, column: 14 - 'numeric value out of range', character: '6'`

According to the [documentation^Wparser](https://github.com/freebsd/pkg/blob/187d1b74fba5f1aa22fcb2e44b828b1ec96dc543/libpkg/pkg_manifest.c#L208-L209), this is an integer or a string. I tested with both `"1.5.2.20.g62e2313"` and `1.5.2.20.g62e2313`, the output is identical. The cirrus CI seems to clone with a depth of 1 and no tags, thus the VERSION_NUM will be `62e2313`.

reynir (approved):

PR #134, opened 2022-11-15
support solo5 0.7.4 option --block-sector-size

hannesm:
//cc @reynir 

missing is a test case where a sector size is specified (and actually run that code).

### DIFF
--- a/command-line/albatross_cli.ml
+++ b/command-line/albatross_cli.ml
@@ -328,15 +328,35 @@ let colon_separated_c =
   and pp ppf (a, b) =
     Fmt.pf ppf "%s:%s" a (match b with None -> a | Some b -> b)
   in
+  parse, pp
+
+let block_c =
+  let parse_block, pp_block = colon_separated_c in
+  let parse s =
+    let ( let* ) = Result.bind in
+    match String.split_on_char '@' s with
+    | [ block ] ->
+      let* (name, device_name) = parse_block block in
+      Ok (name, device_name, None)
+    | [ block; sector_size ] ->
+      let* sector_size =
+        try Ok (int_of_string sector_size)
+        with Failure _ -> Error (`Msg "sector size must be an integer")
+      in
+      let* (name, device_name) = parse_block block in
+      Ok (name, device_name, Some sector_size)
+    | _ -> Error (`Msg "format is 'name[@sector-size]' or 'name:device-name[@sector-size]'")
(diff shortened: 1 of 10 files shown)


src/vmm_resources.ml
```diff
@@ -105,7 +105,7 @@ let use_blocks t name vm active =
   | [] -> t
   | blocks ->
     let block_names =
-      List.map (fun (bd, dev) ->
+      List.map (fun (bd, dev, _sector_size) ->
```
reynir, line 108: Maybe it could be interesting to be able to restrict sector sizes. For another pull request...

reynir (approved):
Looks good! I submitted a small change. I haven't tested it yet. Maybe we could warn against using a too large block size -- but then again I don't know exactly what is "too large".

src/vmm_resources.ml
reynir: I implemented the check in the client in https://github.com/roburio/albatross/pull/134/commits/48cfd63c297202ac6debaeb2e8ac292358e27000

command-line/albatross_cli.ml
```diff
@@ -328,15 +328,40 @@ let colon_separated_c =
...
+      let* (name, device_name) = parse_block block in
+      Ok (name, device_name, Some sector_size)
+    | _ -> Error (`Msg "format is 'name[@sector-size]' or 'name:device-name[@sector-size]'")
+  and pp ppf (a, b, c) =
+    Fmt.pf ppf "%a%a" pp_block (a, b) Fmt.(option ((any "@") ++ int)) c
+  in
   Arg.conv (parse, pp)
 
 let block =
   let doc = "Block device name (block or name:block-device-name)" in
```
reynir, line 360: This description needs to be updated. I will do that in a minute.

reynir:
I tested this branch using opam-mirror:

``` shell
$ albatross-client-local create opam-mirror ./mirror.hvt --block=tar:opam-mirror@4096 --net=service:mirbr0 --mem=256 --arg=--help
# and albatross-client-local console opam-mirror:
console 2022-11-16T10:10:00-00:00: solo5-hvt: Invalid option: `--block-sector-size:tar=4096'
```
Currently I'm waiting for solo5 0.7.4 to go through the builder queue for debian, but it looks good so far. The existing three unikernels came back up just fine after upgrading albatross. In the meantime I will fix the description of `--block`.

PR #137, opened 2022-11-18
Create any missing directories before a doing bind

samoht:


hannesm:
I don't think it is worth it. It complicates the code quite drastically, and suddenly there are `failwith` exceptions deep in code that is executed normally at runtime (where no exceptions are expected). In addition, albatross-console is the daemon to be started first which creates the `console.sock`, so that directory should be ensured to exist by albatross-console.

Maybe you could investigate and provide insight about albatross becoming a system package on your platform?

hannesm:
To elaborate a bit more on that topic:
- the albatross-daemon needs to be executed as superuser for creating virtual machines (at least on FreeBSD, but the systemd service script does the same atm)
- it is crucial that the `tmpdir` can be written to by the albatross user, so that albatross-console and albatross-stats can create sockets there (and also the vmmd.sock is writable by the albatross user)

Now, albatross-daemon doesn't know much about the albatross user (and as mentioned earlier, albatross-console needs to be started first), I'm not sure how this can work out. I'm pretty sure the `albatross` user is not able to consruct file and directories in `tmpdir/albatross/` (or can it on all platforms)?

samoht:
I'm running both daemons as my main user and setting `tmpdir` to `./tmp` - I probably will need superuser rights to create the taps but I'm not there yet. I am just running everything as a process so I don't need superuser rights to create virtual machines.

samoht:
> I don't think it is worth it. It complicates the code quite drastically, and suddenly there are failwith exceptions deep in code that is executed normally at runtime (where no exceptions are expected). In addition, albatross-console is the daemon to be started first which creates the console.sock, so that directory should be ensured to exist by albatross-console.

I agree that failing at runtime is not ideal - maybe I should move this during the daemon startup phase?

For `console.sock`, this should indeed be created by `albatross-console` but the first time I run `albatrossd` it failed, and I was confused why. I thought it was a good idea to give some hints on how to fix this (e.g. start `albatross-console` first)

hannesm:
> For `console.sock`, this should indeed be created by albatross-console but the first time I run albatrossd it failed, and I was confused why. I thought it was a good idea to give some hints on how to fix this (e.g. start albatross-console first)

That sounds like a good idea to improve the error message.

> I agree that failing at runtime is not ideal - maybe I should move this during the daemon startup phase?

That may be a solution, but as I tried to outline, there's some intricate ownership that I believe can't be setup from within albatross-daemon nor albatross-console

> I'm running both daemons under my user account (by just starting them on the command line) and setting tmpdir to `./tmp` - I probably will need superuser rights to create the taps but I'm not there yet. I am just running everything as a process so I don't need superuser rights to create virtual machines.

Ok, as you said, you'll likely need superuser rights for creating tap devices. Maybe not. Maybe there is also a path on Linux and/or FreeBSD to not require superuser privileges which I have not found (yet?).

samoht:
Closing this as it needs a more complete plan first

PR #139, opened 2022-11-23
improve error messages when socket binding fails, and when albatross_…

hannesm:
…daemon fails to connect to console socket

inspired by the report in #137


reynir (approved):

reynir:
Thanks, looks good

PR #141, opened 2022-11-28
debian packaging: add dependencies (gmp, libnl3)

### LINKED ISSUE #140: debian packaging: should libnl-3-dev and libnl-route-3-dev be a dependency?
since albatross-stats depends on these packages, and both ubuntu and debian use the same package names, shouldn't they be part of the dependencies in the packaging/debian subfolder? //cc @reynir

hannesm:
fixes #140 //cc @reynir

### DIFF
--- a/packaging/debian/control
+++ b/packaging/debian/control
@@ -8,6 +8,7 @@ Homepage: https://github.com/roburio/albatross
 Vcs-Browser: https://github.com/roburio/albatross
 Vcs-Git: https://github.com/roburio/albatross.git
 Architecture: all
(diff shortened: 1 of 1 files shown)


reynir:
Ideally we should depend on the non-dev variants. On debian 11 they are named `libnl-route-3-200` and `libnl-route-3-200` respectively. The `-200` part makes me worry the package names are not uniform across the debian/ubuntu variants. I will have to look into that.

I will merge this as this will fix the issue even if it's an overapproximation.

PR #142, opened 2022-11-29
More precise debian dependencies

reynir:
Follow-up of #141.

In [Debian](https://packages.debian.org/search?keywords=libnl&searchon=names&suite=all&section=all) it's called libnl-3-200 from oldoldstable through unstable.

In [Ubuntu](https://packages.ubuntu.com/search?suite=all&searchon=names&keywords=libnl) it's as well called libnl-3-200 from 18.04 through 22.04LTS (and kinetic which I assume is the next release??).

I think it is safe to assume it's called libnl-3-200 in the relevant debian-derived distributions.


hannesm (approved):
looks good to me -- indeed the `-dev` packages shouldn't be needed in the binary, the only thing we're interested in are the shared objects.

Issue #143, opened 2022-12-05
Albatross-tls-endpoint


palainp:
Hi, thanks for your work on albatross, it's very helpful!
I've not updated albatross for a while until last week. I have now an issue starting albatross-tls-endpoint 1.5.3 (the daemon is started with systemd):
```
# albatross-tls-endpoint --version
version v1.5.3 protocol version 5
# albatross-tls-endpoint --port 1025 -v -v -v --verbosity=debug cacert.pem server.pem server.key
albatross-tls-endpoint: internal error, uncaught exception:
                        Unix.Unix_error(Unix.EINVAL, "bind", "")
                        
```

This was the commands I used to have but something may have changed recently. With `strace` I have the following additional information:
```
...
stat("cacert.pem", {st_mode=S_IFREG|0644, st_size=396, ...}) = 0
stat("server.pem", {st_mode=S_IFREG|0644, st_size=518, ...}) = 0
stat("server.key", {st_mode=S_IFREG|0400, st_size=119, ...}) = 0
rt_sigaction(SIGPIPE, {sa_handler=SIG_IGN, sa_mask=[], sa_flags=SA_RESTORER, sa_restorer=0x7ff7a2edb420}, {sa_handler=SIG_DFL, sa_mask=[], sa_flags=0}, 8) = 0
socket(AF_INET6, SOCK_STREAM, IPPROTO_IP) = 5
fcntl(5, F_GETFD)                       = 0
fcntl(5, F_SETFD, FD_CLOEXEC)           = 0
setsockopt(5, SOL_SOCKET, SO_REUSEADDR, [1], 4) = 0
setsockopt(5, SOL_IPV6, IPV6_V6ONLY, [0], 4) = 0
bind(5, {sa_family=AF_INET, sin_port=htons(1025), sin_addr=inet_addr("0.0.0.0")}, 16) = -1 EINVAL (Invalid argument)
write(2, "albatross-tls-endpoint: internal"..., 150albatross-tls-endpoint: internal error, uncaught exception:
                        Unix.Unix_error(Unix.EINVAL, "bind", "")
                        
) = 150
rt_sigaction(SIGCHLD, {sa_handler=0x5652b8785300, sa_mask=[], sa_flags=SA_RESTORER, sa_restorer=0x7ff7a2edb420}, NULL, 8) = 0
[...]

PR #145, opened 2022-12-07
tls-endpoint: allow an ip to listen on being specified on command line

hannesm:
for fixing the other part of #143 (or the suggestion in it)

### DIFF
--- a/src/vmm_lwt.ml
+++ b/src/vmm_lwt.ml
@@ -12,13 +12,13 @@ let safe_close fd =
     (fun () -> Lwt_unix.close fd)
     (fun _ -> Lwt.return_unit)
 
-let port_socket port =
+let port_socket ip port =
   let open Lwt_unix in
   let s = socket PF_INET6 SOCK_STREAM 0 in
   set_close_on_exec s ;
   setsockopt s SO_REUSEADDR true ;
   setsockopt s IPV6_ONLY false ;
-  bind s (ADDR_INET (Unix.inet6_addr_any, port)) >>= fun () ->
+  bind s (ADDR_INET (Ipaddr_unix.to_inet_addr ip, port)) >>= fun () ->
   listen s 10 ;
   Lwt.return s
 
--- a/src/vmm_lwt.ml
+++ b/src/vmm_lwt.ml
@@ -18,6 +18,7 @@ let port_socket ip port =
   set_close_on_exec s ;
   setsockopt s SO_REUSEADDR true ;
   setsockopt s IPV6_ONLY false ;
+  let ip = Ipaddr.V6 (Ipaddr.to_v6 ip) in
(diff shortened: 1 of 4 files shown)


src/vmm_lwt.ml
```diff
@@ -12,13 +12,13 @@ let safe_close fd =
...
 
-let port_socket port =
+let port_socket ip port =
   let open Lwt_unix in
   let s = socket PF_INET6 SOCK_STREAM 0 in
   set_close_on_exec s ;
   setsockopt s SO_REUSEADDR true ;
   setsockopt s IPV6_ONLY false ;
-  bind s (ADDR_INET (Unix.inet6_addr_any, port)) >>= fun () ->
+  bind s (ADDR_INET (Ipaddr_unix.to_inet_addr ip, port)) >>= fun () ->
```
reynir, line 21: Hmm, I worry if we specify `0.0.0.0` that we get the same error on Linux.

reynir: Yes, we do.

hannesm: uhm, so should we change some code? we could use above `PF_INET` if we have an ipv4 address?

reynir, line 21:
```suggestion
  let ip = Ipaddr.V6 (Ipaddr.to_v6 ip) in
  bind s (ADDR_INET (Ipaddr_unix.to_inet_addr ip, port)) >>= fun () ->
```

reynir: I think we can do a dance by converting to ipv6 with `Ipaddr.(V6 (to_v6 ip))` as I just suggested below.

hannesm: 👍🏾

reynir: I just tested binding using the above code with "0.0.0.0", and it doesn't fail, but I don't see it listening in the output from `netstat -tlnp`. I will try to test some more...

reynir: This article suggests that there may be issues with using ::ffff:0.0.0.0 on some platforms: https://lore.kernel.org/all/87tz5s1pxl.fsf@nemi.mork.no/T/

hannesm:
So, what is the opinion on this PR? We could:
- just close it (until someone requests such a feature)
- in the bind have code paths depending on the ip protocol version provided
- just remove albatross_tls_endpoint (and advertise the inetd service more)
- select a single ip protocol version to use

What do you think @reynir? I'm fine with option 2 since it seems to solve the remaining issues.

hannesm:
I propose [a307e58](https://github.com/roburio/albatross/pull/145/commits/a307e588f6a08a9ef6ba5733d3c20785a83b91f7) (part of this PR), and think this can be merged and shouldn't exhibit strange behaviour.

reynir (approved):

PR #146, opened 2022-12-08
host_port: allow multiple ':' in the hostname (i.e. an IPv6 address)

hannesm:


reynir (approved):
I think it is slightly confusing. But it helps that the port is always expected and that the `~docv`s specify `HOST:PORT`.

Maybe we can reconsider the format in a future major point release.

PR #148, opened 2023-01-09
add restart command, split unix parts into a sublibrary

### LINKED ISSUE #118: add a restart command

hannesm:
The first commit fixes #118 

The second commit moves the unix-dependent parts into albatross.unix, and makes the albatross core library independent of unix (and thus usable in a MirageOS unikernel).


reynir (approved):
I think this looks good. Thanks.

PR #151, opened 2023-02-15
Remove logs when we start the albatross_influx daemon

### LINKED ISSUE #150: Don't fill `syslog` with DEBUG logs from `albatross_influx`
Currently, `albatross_influx` runs with `-vv` which populate too fast my log:
https://github.com/roburio/albatross/blob/83dff4a3fe6a4a1486d935550136c59a35645c34/packaging/Linux/albatross_influx.service#L13

It will be nice to just delete this argument when we distribute `albatross` for `systemd`.

dinosaure:
Fix #150

### DIFF
--- a/packaging/Linux/albatross_console.service
+++ b/packaging/Linux/albatross_console.service
@@ -10,7 +10,7 @@ After=syslog.target
 Type=simple
 User=albatross
 Group=albatross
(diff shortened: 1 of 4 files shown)


reynir:
I added a commit to remove verbose logging from the other services on Linux. Let me know if you (don't) agree :-)

dinosaure:
It's perfect!

Issue #149, opened 2023-02-09
document and smooth the CA experience


hannesm:
related is #52 (for adjusting / decreasing the resources for a specific person)

@TheLortex tried to use the remote-tls endpoint and the albatross-client-remote-tls. Noteworthy things:
- albatross-provision-ca sign of an intermediary (sub)CA is allowed even if there are no CPU -- which is slightly useless
- it could as well check there to be at least one VM

Also, the albatross-client-remote-tls takes in a (client) certificate and the CA certificate -- but it expects an entire client CA chain and could locally verify that it's a good certificate before connecting to the remote (and transferring the entire unikernel). The "CA sign" could output a bundle instead of a single certifcate (though it may not know the entire chain... if there's more than one level of delegation involved).

hannesm:
addressed in the mentioned PRs.

PR #155, opened 2023-02-23
Document albatross over tls

TheLortex:
Here is a document I synthesized from my experiments.
It might help for the "document" part of https://github.com/roburio/albatross/issues/149

### DIFF
--- a/TLS.md
+++ b/TLS.md
@@ -70,7 +70,7 @@ albatross-provision-ca sign cacert.pem db ca.key user.req
 | _certificate signing request_ | |            |            | user.req        |        |
 
 5. **client:** the client wants to create an unikernel, instead of using the
-traditional albatross-client-local command, it has to wrap the request in a
+albatross-client-local command, it has to wrap the request in a
 certificate signing request which will be submitted to the intermediate CA.
 
 ```
--- a/TLS.md
+++ b/TLS.md
@@ -1,0 +1,104 @@
+# Albatross over TLS
+
+Albatross uses PKI to authenticate both client and server. Requests are signed
+by a certificate authority (CA) that is trusted by the server. CA can delegate
+resources using policies, which happens by creating an intermediate CA.
+Revokation is not implemented, as delegation happens without the server knowing
+about it.
+
+
+This example shows how one can delegate part of the resources to a user. There
+are 4 entities:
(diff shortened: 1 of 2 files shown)


TLS.md
```diff
@@ -0,0 +1,104 @@
...
+```
+albatross-tls-endpoint cacert.pem server.pem server.key
+```
+
+3. **intermediate CA:** we want to delegate part of the resources to a given
+user. The user generates a signing request to allow a memory of 1024MB to run
+16 unikernels on two CPUs.
+
+```
+albatross-provision-request add_policy user 16 --mem 1024 --cpu 0
```
hannesm, line 49:
```suggestion
albatross-provision-request add_policy user 16 --mem 1024 --cpu 0 --cpu 1
```

In the text, there's `two CPUs` -- that should be reflected in the command.

Also I'm wondering whether we need a slight excursion what these `cpu` arguments are meaning / what they're used for.

TLS.md
```diff
@@ -0,0 +1,104 @@
...
+```
+
+| description                   | | server     |  CA        | intermediate CA | client |
+| ----------------------------- |-| ---------- | ---------- | --------------- | ------ |
+| _private key_                 | | server.key | ca.key     | user.key        |        |
+| _public certificate_          | | server.pem | cacert.pem | **user.pem**    |        |
+| _certificate signing request_ | |            |            | user.req        |        |
+
+5. **client:** the client wants to create an unikernel, instead of using the
+traditional albatross-client-local command, it has to wrap the request in a
```
hannesm, line 73:
```suggestion
albatross-client-local command, it has to wrap the request in a
```

hannesm:
Thanks for writing this up. I added some minor comments.

TLS.md
TheLortex: Good catch, thanks. Indeed I was confused by the `--cpu <n>` option at first I thought `<n>` was the total number of CPUs allowed, instead of the core ID. I'll make the doc more precise on that matter.

TheLortex: I have added a commit to make the doc consistent and replaced `two CPUs` by `CPU IDs 0 and 1`. It still not very clear what that means as I'm not familiar with the option, so don't hesitate to add something if you want !

hannesm:
thanks! :)

PR #157, opened 2023-03-01
Improve ca experience

hannesm:
with this PR, unikernel creation and policy addition are checked before a connection is established / a certificate is generated

addresses some parts of #149 //cc @reynir @TheLortex

### DIFF
--- a/TLS.md
+++ b/TLS.md
@@ -99,6 +99,5 @@ albatross-provision-ca sign user.pem db user.key hello.req
 with the intermediate CA certificate to form the full chain.
 
 ```
-cat hello.pem user.pem > hello.bundle
-albatross-client-remote-tls cacert.pem hello.bundle hello.key <REMOTE_IP:PORT>`
+albatross-client-remote-tls cacert.pem hello.pem hello.key <REMOTE_IP:PORT>`
 ```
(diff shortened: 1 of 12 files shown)


hannesm:
I'm ready to get some review/comments about this. The commits should be understandable one-by-one. What it achieves is to check various things (policy is usable, unikernel matches policy, policy is smaller than the super-policy) locally; it also checks for client-bistro and client-remote-tls that the provided certificate chain validates (before initiating any network communication), and the albatross_provision_ca sign <req> outputs a bundle (thus, the `cat hello.pem user.pem > hello.bundle` is no longer needed).

//cc @reynir @TheLortex

reynir (approved):
Looks good to me. There are some calls to `Lwt.fail_with`, and I'm curious if the exceptions are caught somewhere. If we catch them and print an error that's fine. By using `Lwt.fail_with` over `failwith` we lose the stack trace - and for these exceptions we likely don't want the stack traces.

hannesm:
> Looks good to me. There are some calls to `Lwt.fail_with`, and I'm curious if the exceptions are caught somewhere. If we catch them and print an error that's fine. By using `Lwt.fail_with` over `failwith` we lose the stack trace - and for these exceptions we likely don't want the stack traces.

I agree. The two modules where Lwt.fail_with is used are albatross_client_bistro and albatross_tls_common. I agree this should be revised. I as well think this can be part of this PR (but doesn't need to). To merge this PR, we should maybe test a sample deployment (to ensure there hasn't been any issue (e.g. using `<` instead of `>`)).

reynir:
That sounds like a good way forward. I agree that revising exceptions doesn't have to be a part of this PR.

What we are not checking that we could check are the block commands - is the block size larger than allowed by the policy?

PR #159, opened 2023-03-22
Make albatross-provision one binary

reynir:
`albatross-provision` has subcommands `ca` and `request` that serve the same purpose as `albatross-provision-ca` and
`albatross-provision-request` respectively.

The binary is about the same size as either of the previous binaries. In the (compressed) debian package this results in a further ~3MB reduction of size.


hannesm:
hmm, this leads to quite wordy invocations "albatross-provision ca sign" -- why not flatten that, "albatross-provision sign"?

hannesm:
I took the liberty to rebase and merge the subcommands into `albatross-provision` (as remarked above).

PR #162, opened 2023-04-11
rename albatross-client-local to albatross-local

hannesm:
rename albatross-client-bistro to albatross-remote
 merge albatross-client-remote-tls into albatross-remote as "certificate" subcommand

also get rid of fail_with

once this is in, we should (a) test (b) cut a 2.0 (c) update the documents that refer to albatross commands... what do you think @reynir?


hannesm:
not sure, but albatross-local includes a tls stack etc. for "update" -- so what about merging all that even further into a single `albatross-client` binary which can do local or remote stuff, and may even include the albatross-provision parts
--> all the albatross commands would only exist once (easy to update / reason) --> less code duplication
-> the logic whether to dump to a file (provision), interact with albatross locally (albatross-client-local) or remotely (albatross-client-bistro) would be as a matter of "destination" parameters
-> only some subcommands would be special  (such as inspect-dump / generate-ca / sign-csr / certificate) since they only have a specific use environment.

what do you think?

hannesm:
now everything is merged, this should be good to be reviewed and tested :) //cc @reynir

PR #163, opened 2023-04-25
check tender ABI version against unikernel ABI version

hannesm:
since solo5 0.8.0 is getting released, better check for compatibility before creating the unikernel

### DIFF
--- a/src/vmm_unix.ml
+++ b/src/vmm_unix.ml
@@ -223,14 +223,19 @@ let check_solo5_tender target version =
          `Msg (Fmt.str "tender does not exist, looked for %a"
                  Fmt.(list ~sep:(any ", ") string)
                  (List.map Bos.Cmd.to_string cmds)))
-      (List.fold_left (fun acc cmd ->
+      (List.fold_left (fun acc name ->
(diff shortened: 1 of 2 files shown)


dinosaure:
LGTM

src/vmm_unix.ml
```diff
@@ -209,12 +201,59 @@ type solo5_target = Spt | Hvt
...
+      (fun _ ->
+         `Msg (Fmt.str "tender does not exist, looked for %a"
+                 Fmt.(list ~sep:(any ", ") string)
+                 (List.map Bos.Cmd.to_string cmds)))
+      (List.fold_left (fun acc name ->
+           match acc with
+           | Ok _ as cmd -> cmd
+           | Error _ ->
+             let db_pre = Fpath.(!dbdir / name) in
+             if Bos.OS.File.is_executable db_pre then
```
hannesm, line 231: we can use `Bos.OS.Cmd.must_exist` here, since if this gets the full path passed, it doesn't check for existance neither executability (as documented in the Bos.OS.Cmd moduile).

PR #165, opened 2023-05-25
Nix: Update inputs

Julow:
The package requires a more recent version of `opam-repository`, the lock file has to be updated.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -72,15 +72,11 @@
         "flake-utils": [
           "flake-utils"
         ],
-        "mirage-opam-overlays": [
-          "mirage-opam-overlays"
-        ],
+        "mirage-opam-overlays": "mirage-opam-overlays",
         "nixpkgs": [
           "nixpkgs"
         ],
-        "opam-overlays": [
-          "opam-overlays"
-        ],
+        "opam-overlays": "opam-overlays",
         "opam-repository": [
           "opam-repository"
         ],
@@ -156,10 +152,8 @@
     "root": {
       "inputs": {
         "flake-utils": "flake-utils",
-        "mirage-opam-overlays": "mirage-opam-overlays",
         "nixpkgs": "nixpkgs",
         "opam-nix": "opam-nix",
-        "opam-overlays": "opam-overlays",
         "opam-repository": "opam-repository"
       }
     },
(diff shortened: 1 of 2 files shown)


hannesm:
thanks. two questions:
(1) the nix build fails:
```
error (ignored): error: end of string reached
error:
       … while evaluating the attribute 'defaultPackage'

         at /nix/store/qrsm81kzz217an4wbqh05qr2dcldpkv5-source/flake.nix:40:9:

           39|         packages = { inherit albatross; };
           40|         defaultPackage = albatross;
             |         ^
           41|

       … while evaluating the attribute 'pkgs'

         at /nix/store/j9avmdn9rv5f465s2jc15as8idi43sp9-source/flake.nix:72:18:

           71|         opam-nix = import ./src/opam.nix {
           72|           inherit pkgs opam-repository opam-overlays mirage-opam-overlays;
             |                  ^
           73|         };

       (stack trace truncated; use '--show-trace' to show the full trace)

       error: getting status of '/nix/store/zgdl7i7p1slw1nn8cxl183xsn452kbmp-nixpkgs': No such file or directory
```

(2) why is there "opam-overlays" and "mirage-overlays"? this package should be fine with only the opam-repository.

Julow:
The flake no longer builds with Nix 2.15 for some reason. I'm debugging this. I'm considering pinning version 2.11 in CI as 2.15 is very recent and not everyone can have it easily.

`opam-overlays` and `mirage-overlays` are dependencies of `opam-nix` even though they are only used to be added to the solver input. With some configuration, we could remove them from the solver inputs but they are still marked as dependencies to `opam-nix`.

hannesm:
Thanks for your prompt reply. Sticking to an older nix release sounds fine, as does keeping the overlay repository if "opam-nix" needs them anyways.

Julow:
It has been suggested that we don't need to control the two overlays dependencies in (https://github.com/tweag/opam-nix/issues/43#issuecomment-1563142135) so I'll remove them to clean up `flake.nix`.

hannesm:
thanks a lot

hannesm:
ocaml-ci failures are unrelated ("Cancelled"), and clicking on the url for more details returns an internal server error (500) at the ocaml-ci site... luckily we have github actions & cirrus

PR #166, opened 2023-07-26
Update comments for Linux exotic arches

dra27:
ocaml/ocaml#12405 proposes using `linux` everywhere for the "system" variable. The change doesn't semantically affect the script, but it seemed worth a clarifying comment not to _rely_ on the old system values in future.


hannesm:
Would it make sense to be a bit more specific when you write "Historically" -- i.e. from when till when (since AFAIU the ocaml runtime system will soon change something in that regards?). Thanks a lot for your PR! :)

PR #172, opened 2024-01-09
stats: support FreeBSD 14 (vm_get_stats interface change)

hannesm:


stats/albatross_stats_stubs.c
```diff
@@ -135,7 +135,13 @@ CAMLprim value vmmanage_vmmapi_stats (value octx) {
   const char *desc;
   struct vmctx *ctx = (struct vmctx*)octx;
 
+#if __FreeBSD_version >= 1400000
+  struct vcpu *vcpu;
+  vcpu = vm_vcpu_open(ctx, 0);
```
palainp, line 140: `vm_vcpu_open` allocates memory, shouldn't `vm_vcpu_close` be called somewhere?

hannesm: indeed, thanks for the comment -- does [057463f](https://github.com/robur-coop/albatross/pull/172/commits/057463f0765483be09598be91c6cc547dac8d5f6) address your concerns?

palainp: Yes thanks!

PR #174, opened 2024-05-07
use ohex instead of hex

hannesm:


Julow:
Thanks for pinging. I tested that this builds on top of https://github.com/robur-coop/albatross/pull/175

PR #175, opened 2024-05-08
Nix: Update inputs

Julow:
This updates all inputs at once, including upgrading to the latest opam-repository.


hannesm:
thanks for your prompt reply and PR

PR #177, opened 2024-07-01
Update README.md use correct command

PizieDust:
The command is `add-policy`,
Using `add_policy` gives the error `unknown command 'add_policy', did you mean 'add-policy'?`


hannesm:
Thanks

PR #181, opened 2024-07-01
make console command changes in respect to time backwards compatible

hannesm:
previously, an old client subscribed to a console, but then received generalized_time timestamps (instead of utc_time).

now, this is cleaner: the old client will indicate that they need utc_time timestamp.

discovered by @PizieDust

### DIFF
--- a/daemon/albatross_console.ml
+++ b/daemon/albatross_console.ml
@@ -27,9 +27,15 @@ let read_console id name ring fd =
         Vmm_ring.write ring (t, line) ;
         (match Vmm_core.String_map.find_opt name !active with
          | None -> Lwt.return_unit
-         | Some (version, fd) ->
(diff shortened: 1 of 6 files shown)


hannesm:
This breaks compatibility of stuff currently deployed (from the main branch), but I don't care enough about that.

hannesm:
To be more precise, if you've deployed the latest release, everything is fine.

If you have deployed the current main branch (since Mon Oct 2 10:58:45 2023 5d6e5c4de40c21d7e1f55ebdc1a612ea877f0d96), any 2.0.0 released versions won't interoperate (in respect to console). Now, after this change, again nothing will interoperate. Once you update to after this PR, suddenly 2.0.0 clients and future main branches will work again :)

PR #185, opened 2024-08-12
add support to log via syslog for daemons

### LINKED ISSUE #176: tls-inetd merged with tls-endpoint should be nicer
there's some issues:
- with inetd, log level should be none (since it uses stdin/stdout for communication with the remote peer)
- the checks whether the cert & key are good should be done at startup and leading to a specific exit code
- the documentation (e.g. in the FreeBSD package installation) should be adjusted now that there's no albatross-tls-inetd anymore (and it should be executed with `-q` at the current time)

hannesm:
This is especially useful for albatross-tls-endpoint in inetd mode, where stdout is occupied with client communication

Fixes #176

### DIFF
--- a/daemon/albatross_tls_endpoint.ml
+++ b/daemon/albatross_tls_endpoint.ml
@@ -166,7 +166,7 @@ let ip =
   Arg.(value & opt ip_c Ipaddr.(V6 V6.unspecified) & info [ "ip" ] ~doc)
 
 let inetd =
-  let doc = "Inetd mode" in
+  let doc = "Inetd mode. Be sure to use `--syslog` (or `--quiet` to disable logging)" in
   Arg.(value & flag & info [ "inetd" ] ~doc)
 
 let cmd =
@@ -181,7 +181,7 @@ let cmd =
   ] in
   let term =
     Term.(
-      const jump $ Albatross_cli.setup_log $ cacert $ cert $ key
+      const jump $ (Albatross_cli.setup_log Albatrossd_utils.syslog) $ cacert $ cert $ key
       $ ip
       $ Albatrossd_utils.port_or_socket ~default_port:1025
       $ Albatross_cli.tmpdir
--- a/daemon/albatross_tls_endpoint.ml
+++ b/daemon/albatross_tls_endpoint.ml
@@ -101,7 +101,12 @@ let handle tls =
         (Vmm_commands.header ~version name, reply) >|= fun _ ->
       ()
 
-let jump _ cacert cert priv_key ip port_or_socket tmpdir inetd =
+let jump _ cacert cert priv_key ip port_or_socket tmpdir inetd syslog =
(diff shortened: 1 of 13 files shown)


hannesm:
maybe @reynir has an opinion on this... otherwise I'll merge once CI is fine :)

daemon/albatrossd_utils.ml
```diff
@@ -69,6 +69,10 @@ let systemd_socket_activation =
     let doc = "Pass this flag when systemd socket activation is being used" in
     Arg.(value & flag & info [ "systemd-socket-activation" ] ~doc)
 
+let syslog =
+  let doc = "Pass this flag when syslog should be used for logging" in
```
reynir, line 73: would it be helpful to add at the end: "e.g. when using inetd" ?

daemon/albatross_tls_endpoint.ml
```diff
@@ -166,7 +166,7 @@ let ip =
   Arg.(value & opt ip_c Ipaddr.(V6 V6.unspecified) & info [ "ip" ] ~doc)
 
 let inetd =
-  let doc = "Inetd mode" in
+  let doc = "Inetd mode. Be sure to use `--syslog` (or `--quiet` to disable logging)" in
```
reynir, line 169: Is it perhaps worth it to check this is the case?

daemon/albatrossd_utils.ml
hannesm: addressed in 704d5c23fa94322158ce9d39a6e866242074c80a

daemon/albatross_tls_endpoint.ml
hannesm: Would c12a53f233d0389d549eaf7b84a2853a832870ba be fine in that regard? The issue is that you'll only see exit code 3... since the tls-endpoint binary is only ever called by inetd where stdout (and stderr I believe) are not available anymore...

So, actually I'm not sure whether it is worth to have this check in the code.

reynir: Ah this opens another can of worms. If you pass invalid flags/arguments to albatross-tls-endpoint I think cmdliner will print out help messages?! I think to stderr. If so then that's as well problematic.

reynir: Maybe another approach is to have a `albatross-tls-endpoint-inetd` that is a symlink to `albatross-tls-endpoint` and match on `Sys.argv.(0)` to determine if we're `-inetd` and then pass to `Cmdliner.Cmd.eval` null formatters `~help` and `~err` and set inetd mode.

hannesm: I wouldn't overengineer this. :) Given that for e.g. Linux and systemd we have the socket activation, the inetd mode is mostly a relict for our FreeBSD server :)

PR #188, opened 2024-09-04
require tls 1.0.2 and use the Tls_lwt.read with offset

hannesm:


hannesm:
waits for https://github.com/ocaml/opam-repository/pull/26476

reynir (approved):
Looks good to me!

PR #192, opened 2024-10-04
Fixes: adress renaming (#191), also check in usable policy that block is >= 0

### LINKED ISSUE #191: wording: let's use "unikernel" instead of "vm"
in various pretty-printers and help functions we use the term vm at the moment, while unikernel is more appropriate

hannesm:
- fixes #191
- addresses https://github.com/robur-coop/mollymawk/pull/57#discussion_r1782523357

### DIFF
--- a/src/vmm_core.ml
+++ b/src/vmm_core.ml
@@ -92,7 +92,7 @@ module Name = struct
     path_equal x y && opt_eq (snd x) (snd y)
 
   let pp ppf (p, name) =
-    Fmt.(pf ppf "[vm: %a:%a]" (list ~sep:(any ":") string) p
+    Fmt.(pf ppf "[unikernel: %a:%a]" (list ~sep:(any ":") string) p
            (option ~none:(any "") string) name)
 
   let path (p, _) = p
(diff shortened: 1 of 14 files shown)


PizieDust (approved):
Looks good to me. Thanks @hannesm

src/vmm_core.ml
```diff
@@ -241,21 +241,23 @@ module Policy = struct
...
       Error (`Msg "Unusable policy with no CPUids")
     else if memory <= 16 then
-      Error (`Msg "Unusable policy with memory <= 16 MB")
-    else
-      Ok ()
+      Error (`Msg ("Unusable policy with memory " ^ string_of_int memory ^ " MB <= 16 MB"))
+    else match block with
+      | None -> Ok ()
+      | Some x when x >= 0 -> Ok ()
+      | Some x -> Error (`Msg ("Unusable policy with block " ^ string_of_int x ^ " MB < 0 MB"))
```
reynir, line 260: I'm not sure it's exactly unusable as it would be equivalent to `None`. It's a redundant representation, though. I guess this could break policies made previously?!

reynir: Ah, `Vmm_core.Policy.usable` is only ever called in the client before adding a new policy. So this should not break existing policies.

reynir: Also, I'm not sure this is actually an unusable policy...

hannesm: It's a check that if block is specified, it is > 0. I do think this makes sense, or doesn't it?

reynir: Ah sorry you are right. I somehow read it as x > 0, and and x = 0 makes sense (if redundant)

reynir (approved):

PR #193, opened 2024-10-07
stats: /proc/pid/status parsing more flexible

Firobe:
`/proc/<pid>/status` can have empty values, which were refused by the parsing function. For example on my machine (the separating tabs are still present):
```
x86_Thread_features:	
x86_Thread_features_locked:	
```

Furthermore, and even though no values other than `non/voluntary_ctxt_switches` are used for now, other lines have further leading whitespace (seemingly to align values for humans to read), like `VmPeak`.

Swap to using `String.trim` instead of doing it manually, accounting for all cases


hannesm:
Thanks.

PR #194, opened 2024-10-21
console: Allow for multiple subscribers

reynir:


daemon/albatross_console.ml
```diff
@@ -38,9 +36,20 @@ let read_console id name ring fd =
            Vmm_lwt.write_wire fd (header, data) >>= function
            | Error _ ->
              Vmm_lwt.safe_close fd >|= fun () ->
-             active := Vmm_core.String_map.remove name !active
-           | Ok () -> Lwt.return_unit) >>=
-        loop
+             let update s =
+               let s = Option.value ~default:[] s in
+               let s' = List.filter (fun (_v, _u, fd') -> fd <> fd') s in
```
hannesm, line 41: just to be sure, we can compare file descriptors? I think it is fine

hannesm (approved):
looks good to me

daemon/albatross_console.ml
reynir: I tested with `Unix.stdin = Unix.stdout` and `Unix.stdin = Unix.stdin`. It worked fine in the toplevel. It would be nice to have somewhere to point to that says "it's safe to compare Unix.file_descr` as I remember `Unix.file_descr` being a bit special.

We can also have a counter that we use to generate IDs we compare with.

hannesm: I think it is fine to do. There's been here and there code to take the identifier (integer) of a fd -- but I guess we don't need that nor want to go down that route.

It'd indeed be nice if there was a `Unix.compare_fd` or `Unix.equal_fd` function provided by OCaml.

PR #195, opened 2024-10-24
add block size to unikernel_info, also add the possibility to change arguments in restart

hannesm:
//cc @PizieDust @reynir 

There's still more work needed to retain backwards compatibility.

The main question I have is about the `albatross-client`.. Now the `restart` subcommand can replace the arguments, but this is guarded by an additional flag "--replace-arguments". I came there since just checking whether the arguments are provided or not does not suffice (due to default values, and also maybe a user wants to restart their "hello world" this time with providing no argv at all. -- this wouldn't easily be possible without such a flag. Still it makes me wonder whether we want a separate (sub)command for it -- but then I struggle with naming. Any ideas?

Also, the restart-with-arguments-replaced can't check the manifest on the client side (since there's no image available)... I'm hesitating on what to do here -- I'd like to avoid someone to do a restart and end up with "sorry, your arguments didn't specify the required devices". Any ideas?


hannesm:
I added the needed backwards-compatibility code. This is ready for review.

So, old clients can talk to new servers. There's no attempt to make new clients talk to old servers (they'd need to send old commands, so figure out which version the server has before sending the initial command, or re-connect on failure and try with an older version).

src/vmm_core.mli
```diff
@@ -123,30 +123,56 @@ module Unikernel : sig
...
   val restart_handler : config -> bool
 
+  type arguments = {
+    fail_behaviour : fail_behaviour;
+    cpuid : int ;
+    memory : int ;
+    block_devices : (string * string option * int option) list ;
+    bridges : (string * string option * Macaddr.t option) list ;
+    argv : string list option ;
+  }
```
hannesm, line 133: that's the new record for `restart with arguments`.

src/vmm_core.mli
```diff
@@ -123,30 +123,56 @@ module Unikernel : sig
...
     cpuid : int ;
     memory : int ;
-    block_devices : (string * string option * int option) list ;
-    bridges : (string * string option * Macaddr.t option) list ;
+    block_devices : block_info list ;
+    bridges : net_info list ;
     argv : string list option ;
     digest : string ;
     started : Ptime.t ;
   }
```
hannesm, line 173: I solved some confusion (hopefully) by moving to records for block and network devices -- they also no longer have any optional fields, but are completely filled.

hannesm:
Maybe to elaborate a bit: we're not concerned about API versioning / compatibility here. We just upgrade the API as we want. ;)

What we are concerned are the clients and servers that are deployed by different parties (i.e. not under our control), so esp. an old client should not be punished by our server. In this specific PR, the old client will get reasonable (the same, expected) answers from the new server.

This means that in our serialisation (vmm_asn.ml), we do some extra work to keep track of the old wire formats. In addition the `` `Old_unikernel_info3 `` has been introduced (used to be named `` `Unikernel_info ``), which now results in the `` `Old_unikernels_info3 `` (used to be `` `Unikernel_info ``) -- the wire format of these two constructors is preserved -- the naming differs, since the current one (used by up-to-date clients and servers) is named `` `Unikernel_info ``.

For the restart, it is slightly simpler, since the old command did result in a generic reply - not a specific one (i.e. a success or an error only) -- here we can just translate the old wire format to `` `Unikernel_restart None `` to preserve the existing behaviour.

I guess reading the diff is sometimes confusing (as is writing the code) - and we should add tests for the different versions. But I've been too lazy for writing these tests.

hannesm:
rebased on main, @reynir would you mind to look over this PR?

src/vmm_vmmd.ml
```diff
@@ -370,17 +390,37 @@ let handle_unikernel_cmd t id = function
...
+            { unikernel.Unikernel.config with
+              fail_behaviour = a.fail_behaviour ;
+              cpuid = a.cpuid ;
+              block_devices = a.block_devices ;
+              bridges = a.bridges ;
+              argv = a.argv }
+        in
+        match Vmm_unix.manifest_devices_match ~bridges:config.bridges
+                ~block_devices:config.block_devices
+                config.image
```
reynir, line 420: I think it is preferable that we check the manifest matches before we kill off the existing instance. If they don't match we kill off the existing instance and then don't start a new one.

src/vmm_core.ml
```diff
@@ -375,28 +395,63 @@ module Unikernel = struct
...
     started : Ptime.t ;
   }
 
-  let info t =
+  let info block_size t =
     let cfg = t.config in
+    let block_devices =
+      List.map (fun (unikernel_device, host_device, sector_size) ->
+          let host_device = Option.value ~default:unikernel_device host_device
+          and sector_size = Option.value ~default:512 sector_size (* TODO: this is the default in solo5-hvt at the moment *)
```
reynir, line 441: Possibly a total hack, but we can perhaps match on  the ABI version and use that to determine it's 512. Then we have a migration path if we should want to change the default in solo5 at some point (--> we require an abi change). The downside is I don't think it would be an /actual/ ABI change to change the default...

src/vmm_asn.ml
```diff
@@ -683,18 +762,66 @@ let unikernel_info =
...
+                                   (required ~label:"block-sector-size" int)
+                                   (required ~label:"block-size" int)))))
+         @ (optional ~label:"bridges"
+              (my_explicit 1 (set_of
+                                (sequence3
+                                   (required ~label:"unikernel-device" utf8_string)
+                                   (required ~label:"host-device" utf8_string)
+                                   (required ~label:"mac" mac_addr)))))
+         @ (optional ~label:"arguments"(my_explicit 2 (sequence_of utf8_string)))
+        -@ (optional ~label:"started" (my_explicit 3 generalized_time)))
```
reynir, line 823: Why are these optional?

src/vmm_asn.ml
```diff
@@ -476,6 +476,37 @@ let unikernel_config =
...
+                                   (required ~label:"block-name" utf8_string)
+                                   (optional ~label:"block-device-name" utf8_string)
+                                   (optional ~label:"block-sector-size" int)))))
+         @ (optional ~label:"bridges"
+              (my_explicit 1 (set_of
+                             (sequence3
+                                (required ~label:"netif" utf8_string)
+                                (optional ~label:"bridge" utf8_string)
+                                (optional ~label:"mac" mac_addr)))))
+        -@ (optional ~label:"arguments"(my_explicit 2 (sequence_of utf8_string))))
```
reynir, line 508: Why are these optional?

src/vmm_asn.ml
```diff
@@ -626,17 +667,38 @@ let data =
...
+let old_unikernel_info2 =
   let open Unikernel in
   let f (typ, (fail_behaviour, (cpuid, (memory, (digest, (blocks, (bridges, argv))))))) =
-    let bridges = match bridges with None -> [] | Some xs -> xs
-    and block_devices = match blocks with None -> [] | Some xs -> xs
+    let bridges = match bridges with None -> [] | Some xs ->
+      List.map (fun (unikernel_device, host_device, mac) ->
+          { unikernel_device ;
+            host_device = Option.value ~default:unikernel_device host_device ;
+            mac = Option.value ~default:Macaddr.broadcast mac })
```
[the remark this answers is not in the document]
reynir: Perhaps useful to add a comment why we do that.

src/vmm_asn.ml
```diff
@@ -626,17 +667,38 @@ let data =
...
+        xs
+    and block_devices = match blocks with None -> [] | Some xs ->
+      List.map (fun (unikernel_device, host_device, sector_size) ->
+          { unikernel_device ;
+            host_device = Option.value ~default:unikernel_device host_device ;
+            sector_size = Option.value ~default:512 (* TODO: default from solo5-hvt *) sector_size ;
```
reynir, line 683: Here we don't have the ABI version, so we can't use that.

src/vmm_vmmd.ml
reynir: I will write and push a fix.

src/vmm_asn.ml
reynir: Let's defer that for later.

client/albatross_client.ml
```diff
@@ -693,7 +700,32 @@ let create () force image cpuid memory argv block network compression restart_on
...
+      let* () =
+        if Vmm_core.String_set.(cardinal (of_list (List.map (fun (n, _, _) -> n) block_devices))) = List.length block_devices then
+          Ok ()
+        else
+          Error (`Msg "Block devices must be a set")
+      in
+      let fail_behaviour =
+        let exits = match exit_codes with [] -> None | xs -> Some (Vmm_core.IS.of_list xs) in
+        if restart_on_fail then `Restart exits else `Quit
+      and argv = match argv with [] -> None | xs -> Some xs
```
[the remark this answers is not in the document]
reynir: We can defer that to later.

src/vmm_core.ml
```diff
@@ -375,28 +395,63 @@ module Unikernel = struct
...
+          in
+          let size = Option.value ~default:0 (* TODO: ??? can this happen? *) (block_size host_device) in
```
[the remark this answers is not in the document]
reynir: We can assert false or log an error. I think using 0 is harmless if somehow the block device doesn't exist.

src/vmm_asn.ml
reynir: This is not new.

hannesm: I guess backwards compatibility is the answer. Making things optional means that the old format (without arguments/started) is still a valid thing.

reynir (approved):

reynir:
I will merge #207 then rebase so we can have CI pass.

PR #196, opened 2024-11-20
On solo5 manifest mismatch print the manifest

reynir:
This makes it easier for users of the client to figure out what devices are expected without them having to know to query the solo5 manifest using solo5-elftool themselves. I personally often either get the expected devices wrong or simply don't remember. In that case albatross-client can be a bit more helpful and display the information we have.

The output looks like the following (the indendation is a bit odd):
```
$ albatross-client create teeeest ~/workspace/./dnsvizor/dns-only/dist/dns-stub.hvt
albatross-client: specified bridge(s) does not match with the manifest. Declared manifest: {
                   "type": "solo5.manifest", "version": 1,
                   "devices": [ { "name": "service", "type": "NET_BASIC" } ]
                  }
```

If there's a mismatch of block devices the error message will not say anything about network devices. I'm okay with that.

### DIFF
--- a/src/vmm_unix.ml
+++ b/src/vmm_unix.ml
@@ -252,13 +252,12 @@ let check_solo5_tender target version =
     Error (`Msg (Fmt.str "unexpected solo5 tender --version output, expected one line with 'ABI version %u', got %s"
                    version (String.concat "\n" out)))
 
-let solo5_image_devices image =
-  let* mft = Solo5_elftool.query_manifest (owee_buf_of_str image) in
-  Ok (List.fold_left
-        (fun (block_devices, networks) -> function
-           | Solo5_elftool.Dev_block_basic name -> name :: block_devices, networks
-           | Solo5_elftool.Dev_net_basic name -> block_devices, name :: networks)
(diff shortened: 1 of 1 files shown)


hannesm:
Thanks for your PR. I wonder whether printing the names (of network / block devices) and the names used in the manifest is more concise, and easier to comprehend?

reynir:
Thank you for your comment. I pushed a change which alters the error message:
```
$ albatross-client create teeeest mirror.hvt --block foo --net bar --block quux
albatross-client: specified block device(s) does not match with manifest. Declared devices: net "service", block "tar"
```
I added quotes around the device names. This should not be necessary as solo5 device names are rather restrictive, but I find it makes it more clear what is the device type and what is the name. A somewhat malicious example would be `block "net"` which would read rather confusingly without the quotes.

hannesm:
I do like that. But can we reformulate the "Declared devices", and maybe use "devices oresent in manifest"?

I find the term "declared" confusing... these devices are also declared as boot parameters..

hannesm:
thanks a lot :)

Issue #199, opened 2024-11-20
Out of memory in albatross-console


reynir:
Possibly related to #198 and #194 
```
Nov 20 09:09:13 bark.sorbus.is albatross-console[953]: albatross-console: [ERROR] exception Unix.Unix_error(Unix.EBADF, "check_descriptor", "") while writing
Nov 20 09:09:13 bark.sorbus.is albatross-console[953]: albatross-console: [ERROR] exception Unix.Unix_error(Unix.EBADF, "check_descriptor", "") while writing
Nov 20 09:09:13 bark.sorbus.is albatross-console[953]: albatross-console: [ERROR] exception Unix.Unix_error(Unix.EBADF, "check_descriptor", "") while writing
Nov 20 09:09:13 bark.sorbus.is albatross-console[953]: albatross-console: [ERROR] exception Unix.Unix_error(Unix.EBADF, "check_descriptor", "") while writing
Nov 20 09:09:13 bark.sorbus.is albatross-console[953]: albatross-console: [ERROR] :opam-mirror error while reading Out of memory
```

I will try run the unikernel by hand and see if the output is particularly spammy. I only seem to get one line from albatross-console:
```
console 2024-11-20T09:05:44-00:00:             |      ___|
```
I think this is the first line of the solo5 ASCII logo:
```
                |      ___|
      __|  _ \  |  _ \ __ \
    \__ \ (   | | (   |  ) |
    ____/\___/ _|\___/____/
```

hannesm:
this looks like something is wrong in the data structures... how many console readers do you have/had? 10s or 100s or 1000s?

reynir:
I made the Lwt.catch handler reraise the exception and got this:
```
Nov 20 12:15:39 bark.sorbus.is albatross-console[105714]: Fatal error: exception Out of memory
Nov 20 12:15:39 bark.sorbus.is albatross-console[105714]: Raised by primitive operation at Dune__exe__Albatross_console.read_console.(fun).loop.(fun).f.(fun).update.(fun) in file "daemon/albatross_console.ml", line 41, characters 58-67
Nov 20 12:15:39 bark.sorbus.is albatross-console[105714]: Called from Stdlib__List.find_all.find in file "list.ml", line 242, characters 17-20
[...]

PR #200, opened 2024-11-20
albatross-console: remove closed fds on exceptions

### LINKED ISSUE #198: File descriptor leak in albatross-console
I observed in the syslog that I would get more and more EBADF the more clients I connected and disconnected to albatross-console:

```
Nov 20 09:09:13 bark.sorbus.is albatross-console[953]: albatross-console: [ERROR] exception Unix.Unix_error(Unix.EBADF, "check_descriptor", "") while writing
Nov 20 09:09:13 bark.sorbus.is albatross-console[953]: albatross-console: [ERROR] exception Unix.Unix_error(Unix.EBADF, "check_descriptor", "") while writing
Nov 20 09:09:13 bark.sorbus.is albatross-console[953]: albatross-console: [ERROR] exception Unix.Unix_error(Unix.EBADF, "check_descriptor", "") while writing
Nov 20 09:09:13 bark.sorbus.is albatross-console[953]: albatross-console: [ERROR] exception Unix.Unix_error(Unix.EBADF, "check_descriptor", "") while writing
Nov 20 09:09:13 bark.sorbus.is albatross-console[953]: albatross-console: [ERROR] :opam-mirror error while reading Out of memory
```

Likely introduced in https://github.com/robur-coop/albatross/pull/194.

reynir:
On exceptions we close the file descriptor. Thus we should as well remove it from the list of listeners.

Fixes #198 (although not a "proper" file descriptor leak in the usual sense).

I will look into #199 now and push to this branch.

### DIFF
--- a/daemon/albatross_console.ml
+++ b/daemon/albatross_console.ml
@@ -17,6 +17,14 @@ let pp_unix_error ppf e = Fmt.string ppf (Unix.error_message e)
 let active = ref Vmm_core.String_map.empty
 
 let read_console id name ring fd =
+  let update s =
+    let s = Option.value ~default:[] s in
+    let s' = List.filter (fun (_v, _u, fd') -> fd <> fd') s in
+    if s' = [] then
+      None
+    else
+      Some s'
+  in
   Lwt.catch (fun () ->
       Lwt_unix.wait_read fd >>= fun () ->
       let channel = Lwt_io.of_fd ~mode:Lwt_io.Input fd in
@@ -36,14 +44,6 @@ let read_console id name ring fd =
            Vmm_lwt.write_wire fd (header, data) >>= function
            | Error _ ->
              Vmm_lwt.safe_close fd >|= fun () ->
-             let update s =
-               let s = Option.value ~default:[] s in
-               let s' = List.filter (fun (_v, _u, fd') -> fd <> fd') s in
-               if s' = [] then
-                 None
-               else
-                 Some s'
-             in
              active := Vmm_core.String_map.update name update !active
            | Ok () -> Lwt.return_unit
         in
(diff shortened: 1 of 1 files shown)


daemon/albatross_console.ml
```diff
@@ -17,6 +17,14 @@ let pp_unix_error ppf e = Fmt.string ppf (Unix.error_message e)
 let active = ref Vmm_core.String_map.empty
 
 let read_console id name ring fd =
+  let update s =
```
hannesm, line 20: need to receive the `fd` as well

daemon/albatross_console.ml
```diff
@@ -36,14 +44,6 @@ let read_console id name ring fd =
            Vmm_lwt.write_wire fd (header, data) >>= function
            | Error _ ->
              Vmm_lwt.safe_close fd >|= fun () ->
```
hannesm, line 46: here inside of `f`, we have shadowed the outer `fd`, and what we update below needs to be the inner one.

so, the outer one is the one we read from (i.e. the unikernel output), while the one in `f` is the one from the client, where we write to (and which we need to close on exception / error).

reynir: Oh, you are right! I missed that it was overshadowed.

daemon/albatross_console.ml
```diff
@@ -61,7 +61,8 @@ let read_console id name ring fd =
          | exn ->
            Logs.err (fun m -> m "%s error while reading %s" name (Printexc.to_string exn))
        end ;
-       Vmm_lwt.safe_close fd)
+       Vmm_lwt.safe_close fd >|= fun () ->
+       active := Vmm_core.String_map.update name update !active)
```
hannesm: and now we're at the bottom, we have a huge Lwt.catch and don't know anymore what failed.

we need to introduce one Lwt.catch for the `Lwt_unix.read` from the unikernel fd, and then for each albatross client a separate Lwt.catch to close (and update) the right fd -- but then the albatross-client fds we use Vmm_lwt.write, which should not raise any exceptions -- and thus we shouldn't need a Lwt.catch around it.

Here at the bottom, there was a read error from the unikernel fd, all we can do is close that fd. we don't need to update anything, since the albatross-clients should still get console output, e.g. when the unikernel restarts.

reynir:
Ok, I reverted the original commit (and force pushed because I botched the revert by undoing the second commit).

So I was able to get rid of the exception by using physical equality on the file descriptors. I'm not sure we should install an exception handler in `f` -- it shouldn't raise any exceptions (that we can handle).

hannesm:
looks fine to me

PR #204, opened 2024-12-19
Update README.md

dinosaure:
Fix a typography


dinosaure:
Also, would be nice to precise that `albatross_tls_endpoint` works on `*:1025`.

hannesm:
thanks

PR #206, opened 2025-02-04
albatrossd: on startup, back up state file, fixes #203

### LINKED ISSUE #203: startup: preserve the database
There can be arbitrary complexity / stuff broken when starting albatross daemon:
- bridges may be missing (or being disabled)
- tap interfaces are not set to up_on_open
- whatever else

This leads to albatross starting, but all unikernels failing to start --> sigabort or some other issue with solo5-hvt dumping core. A fix of the system is good, but won't bring us back the unikernels <--> manual intervention needed.

It would be great to first copy the "albatross.state" to a "albatross.state.started" backup file, so we can recover (manually) from such failures, without needing to deal with each unikernel individually.

hannesm:


daemon/albatrossd.ml
```diff
@@ -178,6 +178,11 @@ let jump _ systemd influx tmpdir dbdir =
   match Vmm_vmmd.restore_state () with
```
reynir, line 178: Should we refuse to start if there is a backup state? Otherwise we will overwrite the backup and lose that information.

hannesm: That's a great question.

I thought about having `state.YYYY-MM-DDThh:mm:ss` -- but then we'll just spam the disk full with stuff we're not interested anymore in the soon future.

Then I thought about having two or three state backups (for each start), but I'm undecided about that as well.

So, my current thought is let's have a single backup state. If you start albatross_daemon and notice that some bridges / whatever is missing, you'll notice this early and be happy that you have a backup state. If you don't notice it - well, that's bad luck.

But from the operator perspective, I don't think you'll hammer the albatross_daemon to start 10000 times because some unikernel was not restored. Instead, I'd imagine the operator to look at the log and figure out the reason, and then be happy to find a backup state. WDYT?

reynir:
Rebased on main. Let's merge it when CI passes.

PR #207, opened 2025-02-04
Update to ocaml-solo5-elftool 0.4.0

reynir:
This is pending https://github.com/ocaml/opam-repository/pull/27370

### DIFF
--- a/albatross.opam
+++ b/albatross.opam
@@ -34,6 +34,7 @@ depends: [
   "http-lwt-client" {>= "0.3.0"}
   "happy-eyeballs-lwt"
   "solo5-elftool" {>= "0.4.0"}
+  "cachet" {>= "0.0.2"}
   "fpath" {>= "0.7.3"}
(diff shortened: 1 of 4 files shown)


albatross.opam
```diff
@@ -33,7 +33,7 @@ depends: [
   "ohex" {>= "0.2.0"}
   "http-lwt-client" {>= "0.3.0"}
   "happy-eyeballs-lwt"
-  "solo5-elftool" {>= "0.3"}
+  "solo5-elftool" {>= "0.4.0"}
```
hannesm, line 36:
```suggestion
  "solo5-elftool" {>= "0.4.0"}
  "cachet" {>= "0.0.2"}
```

hannesm:
change looks fine, the FreeBSD CI will continue to fail since it uses the opam.ocaml.org repository, and the sync from the git repo to the webserver takes ages.

if the other CI systems are fine (the nix one fails all the time, we can ignore it), we can merge.

PR #211, opened 2025-04-11
albatross-tls-endpoint: allow the unikernel image of unikernel_create to be the TLS payload

hannesm:
this is step 1 for streaming, and it does not stream from the tls endpoint to albatross -- the main issue is mollymawk, so let's have that stream the user input (or remote image) to albatross -- and albatross for now use the full unikernel image.

I guess with this change we can adapt mollymawk to stream the unikernel image instead of having it completely in memory. //cc @reynir @PizieDust https://github.com/robur-coop/mollymawk/issues/73

### DIFF
--- a/client/albatross_client.ml
+++ b/client/albatross_client.ml
@@ -245,6 +245,7 @@ let policy unikernels memory cpus block bridgesl =
 
 let to_exit_code = function
   | Error `Eof -> Error Success
+  | Error `Tls_eof -> Error Success
   | Error _ -> Error Communication_failed
   | Ok wire -> output_result wire
 
--- a/daemon/albatross_tls_endpoint.ml
+++ b/daemon/albatross_tls_endpoint.ml
@@ -47,11 +47,15 @@ let process fd =
 let read_image tls =
   let rec loop acc =
     Vmm_tls_lwt.read_tls_chunk tls >>= function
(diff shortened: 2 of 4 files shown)


hannesm:
But reading further, I guess we cannot provide the length of the binary, since that isn't known upfront. We'll need some way to indicate the end of the image.

We can use a protocol doing chunking in the form of:
- the certificate carries unikernel_create but no image
- on the TLS stream, what is sent is <0x00112233> <- 4 byte length, followed by data; rinse, repeat -- until the 4 byte length field is all 0 (marking the end of the data)

reynir (approved):

hannesm:
I pushed a commit for block_add / block_set streaming.

This is (mostly) ready to be merged, but before a release we should revise the internals for block_set / block_add / block_dump -- since at the moment albatross requires for these to have everything in memory, which won't work if we have a huge block device. so, we'll need to push the streaming one level further (and also deal with the local unix communication and streaming).

For unikernel images (Unikernel_create), I'm not worried -- they need to fit into memory in any case.

reynir:
> we should revise the internals for block_set / block_add / block_dump

As far as I'm concerned we can defer that to another PR. With this we are not worse off.

hannesm:
sounds good to me, should we merge then?

reynir:
Thanks!

PR #213, opened 2025-06-04
Stream the block_dump data while reading the contents

hannesm:
Now, the reply to `` `Block_dump `` is an empty piece of data, but the contents thereof is streamed from albatross.

This retains the TODO to re-implement compression, maybe @dinosaure could take a look.

Similar changes (but the other way around, i.e. the data is streamed to albatross daemon [not from albatross daemon]) are needed for `` `Block_set `` and `` `Block_add `` (the latter if `data` is provided).

While looking at it, the `` `Block_set `` / `` `Block_add `` is not fault-safe -- if we cut the power in the middle, the on-disk state will be bad. I'll address this thereafter.

### DIFF
--- a/client/albatross_client.ml
+++ b/client/albatross_client.ml
@@ -73,11 +73,11 @@ let output_result state ((hdr, reply) as wire) =
               write_to_file name cfg.compressed cfg.image)
           unikernels;
         Ok `End
-      | `Block_device_image (_compressed, "") ->
+      | `Block_device_image (compressed, "") ->
         (* TODO: compression! *)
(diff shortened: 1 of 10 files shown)


reynir (changes requested):

hannesm:
@reynir unfortunately your last commit "client: pass around compression state" doesn't compile. I'll leave this PR as is, but continue on #216

hannesm:
@reynir would you be fine to rebase and continue on #216 instead - and we can close this PR?

hannesm:
superseded by #216

PR #214, opened 2025-06-04
Thread state in client

reynir:
This threads state in the client so we can have an open file descriptor to dump a block device to. It also makes the protocol slightly more strict in the client. We could make it even more strict.

This lacks safeguards to close the `` `Dump_to fd `` file descriptor in case of errors.

I'm not sure I like this change.

### DIFF
--- a/client/albatross_client.ml
+++ b/client/albatross_client.ml
@@ -31,7 +31,10 @@ let classify_tls_error = function
     Communication_failed
 
 let exit_status = function
-  | Ok () -> Ok Success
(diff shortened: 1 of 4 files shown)


hannesm:
I like it very much. About safeguards, in the command-line client I'm fine with it throwing exceptions and whatnot -- in other cases (server/daemon) I'd be more concerned about potential file descriptor leaks. The question is whether we should use (more advanced/complex) types for mitigating the "impossible" situations... and I don't think it is really worth it here.

Issue #178, opened 2024-07-01
nix update opam-repository


hannesm:
Dear @Julow, would you mind to do the magic so that the nix stuff uses a more up-to-date opam-repository? The current CI fails since it uses an opam-repository that points to unavailable tarballs (due to git.robur.coop being not available). Thanks a lot.

Julow:
Hi, thanks for pinging and sorry that the nix files are causing problems.

I've written down the commands needed to update the inputs in flake.nix in https://github.com/robur-coop/albatross/pull/179 but that requires having Nix installed.
I think it should be possible to make a Github Action that do that when trigerred manually from the Actions tab. What do you think ?

hannesm:
> I've written down the commands needed to update the inputs in flake.nix in #179 but that requires having Nix installed. I think it should be possible to make a Github Action that do that when trigerred manually from the Actions tab. What do you think ?

Thanks a log. Indeed, a Github action for doing this would be great!

PR #215, opened 2025-06-04
remove nix CI

hannesm:
The underlying reason is that there hasn't been enough interest in the NixOS CI system, and there was nobody caring enough to write documentation on how to update the "flake" stuff (on a non-NixOS machine, or via a GitHub action).


hannesm:
I'll merge. If there's again need for NixOS support, we can revive this CI.

hannesm:
@Julow of course we can re-add them. But I thought we'd need to update some hashes in there if albatross requires newer packages from opam-repository? If I understand from https://github.com/robur-coop/albatross/issues/178, there was the plan to have a GitHub action that updates these hashes!? -- If that is possible, I'd be happy to get that in place (and get the CI back in action).

Having the flake.nix and flake.lock with outdated opam-repository hashes will likely fail silently, thus I'd prefer to have a CI (which now that it has been failing for months wasn't of much use either).

Julow:
Sorry for the slow response on https://github.com/robur-coop/albatross/issues/178, I've totally forgot about it.
Can I suggest adding them back *if* I manage to implement the CI too ?

> Having the flake.nix and flake.lock with outdated opam-repository hashes will likely fail silently,

Absolutely, but users can override the hashes on their side, which they might do anyway for keeping everything uptodate and use the same opam-repository version on their own packages. Coming up with the description that was in flake.nix a lot harder.

hannesm:
> Sorry for the slow response on #178, I've totally forgot about it. Can I suggest adding them back _if_ I manage to implement the CI too ?
> 
> > Having the flake.nix and flake.lock with outdated opam-repository hashes will likely fail silently,
> 
> Absolutely, but users can override the hashes on their side, which they might do anyway for keeping everything uptodate and use the same opam-repository version on their own packages. Coming up with the description that was in flake.nix a lot harder.

Sure. Let's take a first step and revert the removal, #217. I'm happy to get updates from your side (which may be either updating the hashes now, or doing the work to fix #178).

Julow:
Thanks! I updated the inputs here: https://github.com/robur-coop/albatross/pull/218

PR #216, opened 2025-06-04
Stream block_add and block_set

hannesm:
this adds the streaming, to both client and tls_endpoint

### DIFF
--- a/src/vmm_lwt.ml
+++ b/src/vmm_lwt.ml
@@ -248,8 +248,13 @@ let uncompress_stream input =
        else Lwt.return_unit) >|= fun () ->
       stream#close
     | `Malformed err ->
-      stream#push (`Malformed err) >|= fun () ->
-      stream#close
+      stream#push (`Malformed err) >>= fun () ->
+      begin
+        (* This is needed so we can drain [input] *)
+        Lwt_stream.get input >>= function
+        | Some _ -> loop zl rem
+        | None -> stream#close; Lwt.return_unit
+      end
     | `Await zl when is_it_anything_left rem ->
       let (str, src_off, str_len) = Option.get rem in
       let len = Int.min str_len (Bigstringaf.length i) in


dinosaure:
About the compression, we should probably can just `Lwt_stream.map` the given stream instead of a use of `Lwt.async`.

reynir:
I thought about that, but I'm not sure we have a guaranteed one-to-one correspondence. I mean, we may get a string that is not long enough that we can output something?

src/vmm_lwt.ml
```diff
@@ -173,3 +173,75 @@ let write_wire s wire =
...
+        let off = Zl.Def.src_rem zl in
+        Bigstringaf.blit_from_string data ~src_off:0 i ~dst_off:off ~len:(String.length data);
+        let zl = Zl.Def.src zl i 0 (off + String.length data) in
+        loop true zl
+      | None ->
+        let off = Zl.Def.src_rem zl in
+        let zl = Zl.Def.src zl i off 0 in
+        loop true zl
+  in
+  r, loop false zl
```
dinosaure, line 212: A note about `decompress`, it expects to already have an output buffer **before** to call `Zl.Def.encode`. You should write:
```ocaml
loop false (Zl.Def.dst zl o 0 (Bigstringaf.length o))
```

src/vmm_lwt.ml
```diff
@@ -173,3 +173,75 @@ let write_wire s wire =
...
+      let len = Bigstringaf.length o - Zl.Def.dst_rem zl in
+      (if len > 0 then
+         stream#push (Bigstringaf.substring o ~off:0 ~len)
+       else Lwt.return_unit) >|= fun () ->
+      stream#close
+    | `Await zl ->
+      Lwt_stream.get input >>= function
+      | Some data ->
+        (* XXX(reynir): what if [data] doesn't fit?! *)
+        let off = Zl.Def.src_rem zl in
```
dinosaure, line 203: I think you misunderstand `src_rem`. Actually, when `decompress` emits `Await`, all the input buffer is completely free. You can just `blit` the given `str` to `~dst_off:0 ~len:(String.length data)`.

src/vmm_lwt.ml
```diff
@@ -173,3 +173,75 @@ let write_wire s wire =
...
+    | `End zl ->
+      let len = Bigstringaf.length o - Zl.Def.dst_rem zl in
+      (if len > 0 then
+         stream#push (Bigstringaf.substring o ~off:0 ~len)
+       else Lwt.return_unit) >|= fun () ->
+      stream#close
+    | `Await zl ->
+      Lwt_stream.get input >>= function
+      | Some data ->
+        (* XXX(reynir): what if [data] doesn't fit?! *)
```
dinosaure, line 202: Another way around to solve this situation is to make an inner loop which tries to consume `(data, off (* initialy to 0 *), len (* initally to String.length data *))` and execute the loop until `len = 0`.

src/vmm_lwt.ml
```diff
@@ -173,3 +173,75 @@ let write_wire s wire =
...
+  let w = De.Lz77.make_window ~bits:15 in
+  let q = De.Queue.create 0x1000 in
+  let zl = Zl.Def.encoder `Manual `Manual ~q ~w ~level in
+  let i = Bigstringaf.create De.io_buffer_size in
+  let o = Bigstringaf.create De.io_buffer_size in
+  let r, stream = Lwt_stream.create_bounded 2 in
+  let rec loop aw zl =
+    match Zl.Def.encode zl with
+    | `Flush zl ->
+      if aw then begin
```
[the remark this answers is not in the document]
hannesm: see above, I suspect your fix makes this workaround bf6dfdb7183694fa68a23b0a3fd2164af33e86cd superfluous

hannesm:
When looking into the diff (and diff between the latest release), we introduce a different block_dump command, so old clients can talk to new servers (using the old protocol); but there's no new block_add/block_set/unikernel_(force)_create -- so new clients (which use streaming there) won't get a nice error message from old servers.

New servers are still able to interoperate with old clients, since it is always dispatched on the empty (optional) string being present, and the code path for the full block device data being part of the command is still around.

hannesm:
I went ahead and created custom new commands for block_add/block_set and also for the block_dump reply. now we don't overload the "this command may carry an (optional) empty string.

I didn't test the code yet.

client/albatross_client.ml
```diff
@@ -457,6 +457,13 @@ let gen_cert (cert, certs, key) key_type name (cmd : Vmm_commands.t) =
       (fun e -> `Msg (Fmt.to_to_string X509.Validation.pp_signature_error e))
       (X509.Signing_request.sign csr ~valid_from ~valid_until ~extensions key issuer)
   in
+  let* () =
+    let encoded = X509.Certificate.encode_der mycert in
+    if String.length encoded > 1 lsl 24 then
+      Error (`Msg "certificate too big for TLS")
```
reynir, line 463: When does this happen? When the unikernel is too big, or if the block data is too big, I guess. Do we have a backup method if this fails? I think I remember talk about streaming the unikernel image / block data in the TLS tunnel. I guess this is not a blocker for this PR though.

src/vmm_unix.ml
```diff
@@ -486,12 +495,111 @@ let dump_block name =
...
+    (fun e ->
+       Logs.err (fun m -> m "error writing %a: %s"
+                    Fpath.pp name (Printexc.to_string e));
+       safe_close fd >>= fun () ->
+       let rec drop_stream s =
+         Lwt_stream.get s >>= function
+         | None -> Lwt.return_unit
+         | Some _ -> drop_stream s
+       in
+       drop_stream stream >|= fun () ->
```
reynir, line 582: Here we should cancel the reader I think :/

reynir: Or maybe it's fine - the other end will not expect a message until they've sent all their data.

reynir (approved):
I think this mostly looks good. I think it could have been nice to keep create-block with data as a single command: because then we can undo the block if we fail transferring all the data. However, propagating the error back so we can undo the block seems quite hairy with the current code. So let's look at that in the future (maybe when we have `albatross-block`?)

src/vmm_unix.ml
reynir: Ahhh hmm I'm now not sure this works if the compressed data is malformed. In that case we close the uncompressed stream, and then we can't consume the remainder of the underlying compressed stream.

client/albatross_client.ml
hannesm: this happens when the block data/unikernel data is bigger than 2^24, and thus transmitting it over tls leads to truncation.

it is not clear to me whether this is a good spot to put it, esp. with this PR being merged we won't have such big certificates anymore (since all is streamed). I developed this patch for the 2.4.1 release where such a ibg block device would simply fail with a strange error...

reynir: Thanks, that makes sense. I marked this comment chain as resolved.

PR #222, opened 2025-08-15
packaging: register libev as dependency

hannesm:


hannesm:
approved & CI is green, I'll merge

PR #223, opened 2025-08-15
albatross-client: fix console reading, do not close the connection immediately

hannesm:
we receive a `` `sucess `` and then expect actual console data to flow over the connection. if the client closes the connection after receiving the success, we never see console data.

### DIFF
--- a/client/albatross_client.ml
+++ b/client/albatross_client.ml
@@ -96,7 +96,7 @@ let output_result state ((hdr, reply) as wire) =
         let name = hdr.Vmm_commands.name in
         write_to_file name compressed image;
         Lwt.return (Ok `End)
-      | _ -> Lwt.return (Ok `End)
+      | _ -> Lwt.return (Ok state)
     end
   | `Data `Block_data None ->
     (match state with


client/albatross_client.ml
```diff
@@ -96,7 +96,7 @@ let output_result state ((hdr, reply) as wire) =
         let name = hdr.Vmm_commands.name in
         write_to_file name compressed image;
         Lwt.return (Ok `End)
-      | _ -> Lwt.return (Ok `End)
+      | _ -> Lwt.return (Ok state)
```
reynir, line 99: Hmm I don't quite understand. If I get Merlin to expand the match I get these constructors:
```suggestion
      | `Block_devices _ | `Empty | `String _ | `Old_unikernel_info3 _
      | `Old_unikernel_info2 _ | `Unikernel_info _ | `Policies _ ->
        Lwt.return (Ok state)

```
Ah, and then I find the relevant code: https://github.com/robur-coop/albatross/blob/e77ab60aa3dace0375a07be34f03495466fe4534/daemon/albatross_console.ml#L216-L227
We send `` `Success `String "..." ``.

reynir: What I'm unsure of is if we transition `` `Single `` state into `` `End ``...

hannesm: The value `state` is coming as input to this function. It is what in other places is called `next`, and the (second projction of the) return value of `Vmm_commands.endpoint`.

Now, for Console output, the console daemon replies immediately with `` `Success "subscribed" ``, and then keeps the stream open to send the console messages over.

In the current code, we receive and process the `` `Success `` and then tell our caller (the loop in `read`, which calls `p` (which may be `process_local` or `process_remote`)) to `` `End `` -- i.e. to terminate the connection. But that is not what we want to do, but we want to continue to read the stream to receive and see the console messages.

reynir: Yes, I understand. The case when we send `` `Unikernel_cmd `Unikernel_info `` we have state `` `Single `` because we expect only one reply. So after we have read one reply we should transition the state to `` `End `` **iff** in `` `Single `` state. Otherwise we continue to be in `` `Single `` state after reading one reply. So this change makes it so for `` `Unikernel_cmd `Unikernel_info `` we keep expecting "just one more" reply from the daemon, but thankfully the daemon seems to close the connection after sending the reply.

I pushed a commit that I think addresses that.

PR #224, opened 2025-08-15
albatross-client: new subcommand to extract a single unikernel from a dump

hannesm:


client/albatross_client.ml
```diff
@@ -967,6 +967,38 @@ let inspect_dump _ name dbdir =
...
+        Logs.err (fun m -> m "unikernel not found");
+        Cli_failed
+      | Some u ->
+        let filename = filename unikernel_name in
+        let image =
+          if u.Vmm_core.Unikernel.compressed then
+            Result.get_ok (Vmm_compress.uncompress u.image)
+          else
+            u.image
+        in
```
reynir, line 995: We *could* use the streaming uncompress here, but I don't think it's worth it.

client/albatross_client.ml
```diff
@@ -1579,6 +1611,17 @@ let inspect_dump_cmd =
   in
   Cmd.v info term
 
+let extract_dump_cmd =
+  let doc = "Extracts a unikernel from an albatross dump file." in
+  let man =
+    [`S "DESCRIPTION";
+     `P "Extracts a unikernel from an albatross dump file."]
```
reynir, line 1618:
```suggestion
     `P "Extracts a unikernel from an albatross dump file. \
         The unikernel configuration is printed to stdout and the image is dumped to disk."]
```

reynir (approved):
Looks good to me! Thanks

Issue #53, opened 2020-12-01
startup dependencies


hannesm:
since albatross dumps all unikernels to disk, and restarts them on startup, it'd be nice to have control over the start ordering.

using names here would be quite a lot of hassle, but I think a single priority (similar to nice values, or your old /etc/rc.d/SYY-foo numbers): start in ascending order, by default put all unikernels at 100.

with this feature, a DNS resolver & DHCP server can be started early by setting its priority to a lower number.

hannesm:
done in #225

Issue #221, opened 2025-08-12
revise albatross protocol (backwards incompatible)


hannesm:
revise the protocol, and remove some old commands:
- bump the protocol version
- remove the serial number
- remove old-console-subscribe
- remove old_unikernel_info* old_unikernel_get
- remove old_block_dump
- merge unikernel_create and unikernel_force_create into a single command, add the force flag to Unikernel.config
- remove the `string option` from block_add (also the bool), add a convenience in the client to create a block device and fill with data
- remove the string argument from block_set (use the stream instead)
- remove the console_data, use utc_console_data instead
- remove the old_unikernels and old_unikernel_info*
- remove block_device_image

hannesm:
we now revised the stat and have `` `Old_stats_subscribe / `Old_stats_data `` that should be removed one day.

I as well think that our naming can be tighter -- all we need is two strings, so not a chain of prefixes but only a single one. This will make the code easier, remove some use cases that haven't been used in the past -- and make integration with mollymawk more straightforward.

hannesm:
the stats stuff has again be revisited in #290 -- and while I first wrote backwards compatibility code, I deleted that. esp. for stats it is quite some code, plus this is all in memory data anyways, so I went the lazy path and deleted the code for backwards compatibility.

PR #225, opened 2025-08-22
add "startup" to the unikernel config, the lowest is started first

hannesm:
this addresses #53

Please note, since we use "optional" in the ASN.1 at a place where it is not ambiguous, we don't need to revise the wire format. If we'd have put it before/after another "int" field, it would be ambiguous.

### DIFF
--- a/client/albatross_client.ml
+++ b/client/albatross_client.ml
@@ -186,11 +186,11 @@ let http_get_binary ~happy_eyeballs host job build =
 
 let prepare_update ~happy_eyeballs level host dryrun = function
   | Ok (_hdr, `Success (`Unikernel_info
-      [ _name, Vmm_core.Unikernel.{ digest ; bridges ; block_devices ; argv ; cpuid ; memory ; fail_behaviour ; typ = `Solo5 as typ ; _ } ]))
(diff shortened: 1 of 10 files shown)


hannesm:
hmm, but it may be worth to nevertheless bump the wire structures -- otherwise the new client will miscommunicate with an old server. and a new server will send a reply that an old client doesn't understand.

reynir (approved):
Looks good to me

reynir:
> hmm, but it may be worth to nevertheless bump the wire structures -- otherwise the new client will miscommunicate with an old server. and a new server will send a reply that an old client doesn't understand.

Yes, we have otherwise been careful to have compatibility so that's a good point

hannesm:
I pushed a commit with all the new wire stuff, and the server handling old clients gracefully.

I propose really soon (tm) to walk over #221 to get rid of a bunch of code. the only backwards compatibility we should care about is the on-disk data (state_of_str, of_cert_extension). from once 221 is done on, we can as well plan & document how to interoperate (a) new client, old server (b) old client, new server (where (a) is at the moment not in scope, would need some code in albatross-client to try with an older command if the (old) server replied with a (decoding) failure as reply to the new (unknown) command.

PR #229, opened 2025-09-10
Add --name to the unikernel arguments

hannesm:
This is what mirage 4.10.0 allows :D and since albatross already has a name for a unikernel. let's pass that on so we have a unique identifier both inside and outside the unikernel.


hannesm:
I tested it, and by default I think the path shouldn't appear. We've to rethink at some point how a setup should look, and whether we want to share a Grafana & syslog service?

PR #233, opened 2025-10-06
Vmm_core.Name: add a type label, and string conversion, and a compare function

hannesm:


hannesm:
I revised and introduced some submodules //cc @PizieDust @reynir

src/vmm_core.mli
```diff
@@ -22,48 +22,67 @@ module IM : sig
 end
 
 module Name : sig
-  type t
+  module Label : sig
+    type t
+
+    val compare : t -> t -> int
+    val equal : t -> t -> bool
+    val is_empty : t -> bool
```
hannesm: this binding and the one below, will have to disappear. there's one use in vmm_trie which we can replace by option.

src/vmm_core.mli
```diff
@@ -22,48 +22,67 @@ module IM : sig
...
+
+    val to_labels : t -> Label.t list
+
+    val root : t
+    val is_root : t -> bool
+    val parent : t -> t
 
-  type path
+    val append : t -> string -> (t, [> `Msg of string ]) result
+    val append_exn : t -> string -> t
```
hannesm: not sure whether we need this anymore, since you've Label.of_string_exn and append_label

src/vmm_core.ml
```diff
@@ -99,19 +160,10 @@ module Name = struct
...
   let create_of_path p = (p, None)
 
-  let create_exn path name =
-    match create path name with
-    | Ok t -> t
-    | Error `Msg m -> invalid_arg m
-
   let drop_path (_, name) = [], name
 
   let rec drop_prefix_exn (p, name) p' =
```
reynir: I see this function wasn't touched, but I notice it uses `String.equal` instead of `Label.equal` and that's likely too strict.

src/vmm_core.ml
```diff
@@ -140,23 +184,17 @@ module Name = struct
...
-    let ps = match ps with | ""::tl -> tl | ps -> ps in
-    if List.for_all valid_label ps then
-      Ok ps
-    else
-      Error (`Msg "invalid path")
-
   let to_string (p, n) =
-    path_to_string p ^ ":" ^ Option.value ~default:"" n
+    Path.to_string p ^ ":" ^
+    Option.value ~default:"" (Option.map Label.to_string  n)
```
reynir: I think here it's not necessary to do `Label.to_string` because `Label.t` isn't opaque here. But it's fine.

src/vmm_core.ml
```diff
@@ -54,42 +54,103 @@ module Name = struct
...
+      String.get s 0 <> '-' && (* leading may not be '-' *)
+      String.for_all (function
+          | 'a'..'z' | 'A'..'Z' | '0'..'9' | '-' | '.' -> true
+          | _ -> false)
+        s (* only LDH (letters, digits, hyphen, period)! *)
+
+    let of_string str =
+      if valid str then
+        Ok str
+      else
```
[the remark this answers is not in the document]
reynir: I tend to agree. We have a trie so we can expect quite a few calls to `compare`.

src/vmm_core.mli
reynir: Yes, I would drop both `append` and `append_exn` as it simplifies the `Path` code. We can consider adding `Label.of_string_exn` though.

src/vmm_core.ml
```diff
@@ -54,42 +54,77 @@ module Name = struct
...
+    type t = string
+
+    let [@inline always] valid s =
+      String.length s < 64 &&
+      String.length s > 0 &&
+      String.get s 0 <> '-' && (* leading may not be '-' *)
+      String.for_all (function
+          | 'a'..'z' | 'A'..'Z' | '0'..'9' | '-' | '.' -> true
+          | _ -> false)
+        s (* only LDH (letters, digits, hyphen, period)! *)
```
hannesm: with todays view, this restriction is a bit arbitrary. we may want our unikernels to be named 🛩️ or 😄 or 🍌... but we can wait a bit and figure out later (I'm not sure what other systems we interact with want as names..)

src/vmm_trie.ml
```diff
@@ -1,10 +1,11 @@
 (* (c) 2018 Hannes Mehnert, all rights reserved *)
 
+module Map = Map.Make(Vmm_core.Name.Label)
 (* each node may have a value (of type 'a), the boolean represents whether it
    is a path or a name (i.e. foo:bar: <value> or foo:bar <value>). *)
-type 'a t = N of ('a * bool) option * 'a t Vmm_core.String_map.t
```
hannesm: todo: check whether the String_map is still used somewhere, otherwise remove it from Vmm_core

src/vmm_core.mli
```diff
@@ -22,48 +22,63 @@ module IM : sig
...
   val pp : t Fmt.t
 
-  val valid_label : string -> bool
+  val path : t -> Path.t
+  val name : t -> Label.t option
 
-  val path : t -> path
-  val name : t -> string option
+  val create : Path.t -> Label.t -> t
+  val create_of_path : Path.t -> t
```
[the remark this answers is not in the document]
hannesm: done in 9ac544dd79b5c7947a8127e6cbdcad6137c2c8a9

src/vmm_trie.ml
```diff
@@ -14,61 +15,62 @@ let insert id e t =
...
+  go t (Vmm_core.Name.to_labels id)
 
 let append_name prefix name =
   let path =
     let pre_path = Vmm_core.Name.path prefix in
     Option.fold
       ~none:pre_path
-      ~some:(fun prefix_name -> Vmm_core.Name.append_path_exn pre_path prefix_name)
+      ~some:(fun prefix_name ->
+          Vmm_core.Name.Path.append_label pre_path prefix_name)
```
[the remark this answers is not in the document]
hannesm: done in 34d13eee4d8a883b51957be8dd86aae82672e8ee

src/vmm_trie.ml
hannesm: And thanks for looking into this :)

hannesm:
IMHO this is ready. Could be interesting to adapt the current mollymawk with these changes, and see whether there's something missing.

Issue #237, opened 2025-11-10
albatross-console memory usage, take 2


hannesm:
from https://github.com/robur-coop/mollymawk/issues/168

> logs are still available from albatross-console

I think this is actually something we need to revise... Since it opens an opportunity to use lots of memory in albatross_console that is never released: an attacker can spawn fast-console-outputting and dying unikernels, and this is never released.

Instead, we should have per domain/path a LRU of (n + 1) entries where n is the number of unikernels allowed by that domain, thus: if you may spawn 2 unikernels, you get logs of up to 3 unikernels.

This will then make the memory usage of albatross-console smaller. it requires policies (and policy updates) to be communicated.

At the same time, we can have a new command to list/enumerate all available console outputs -- and then mollymawk can use that to figure out which no-longer-running unikernels exist. Note this is kept separate from the unikernel_info stuff, since console doesn't persist anything on disk.

PR #238, opened 2025-11-10
albatross-console: revise, reduce memory footprint

hannesm:
addresses most of #237

### DIFF
--- a/daemon/albatross_console.ml
+++ b/daemon/albatross_console.ml
@@ -34,7 +34,7 @@ end
    (version and file descriptor), a boolean field whether this console is active
    (the unikernel is running), and a ringbuffer (maximum 1024 entries). *)
 type console_map =
-  ((Vmm_commands.version * Lwt_unix.file_descr) list * bool * string Vmm_ring.t) LMap.t
+  ((Vmm_commands.version * Lwt_unix.file_descr) list * bool * string Vmm_ring.t option) LMap.t
 
 (* The global state: a trie, where the key is the path, and the value is pair of
    "number of unikernels allowed on this path" and "console map". *)
@@ -155,7 +155,7 @@ let add_fifo n (path, lbl) =
     (match Trie.find path !state with
      | None ->
        let ringbuffer = Vmm_ring.create "" () in
-       let map = LMap.singleton lbl ([], true, ringbuffer) in
+       let map = LMap.singleton lbl ([], true, Some ringbuffer) in
        let trie, _ = Trie.insert path (n, map) !state in
        state := trie;
        Lwt.return ringbuffer
@@ -163,7 +163,7 @@ let add_fifo n (path, lbl) =
        match LMap.find_opt lbl map with
        | None ->
          let ringbuffer = Vmm_ring.create "" () in
-         let map = LMap.add lbl ([], true, ringbuffer) map in
+         let map = LMap.add lbl ([], true, Some ringbuffer) map in
          (if LMap.cardinal map > n then
             match LMap.choose_opt (LMap.filter (fun _ (_, active, _) -> not active) map) with
             | Some (key, (fds, _, _)) ->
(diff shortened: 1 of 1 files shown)


hannesm:
this now implements the first step of #237, and includes in the `console_add` command the number of unikernels allowed on that path.

a next change is to revise the storage model in albtross-console to adhere to the numbers (and release memory)

hannesm:
There are some pieces missing: the subscribers data structure is not yet reducing.

Also, I'm no longer sure about the LRU, since it's likely not what we want... there can be 3 running unikernels, where 1 is not outputting anything. Now the two others are outputting data onto console (and thus get promoted). When now the two unikernels are destroyed, and new ones are created, the active (still running, silent) one will be discarded by the LRU.

--> we instead can go a simpler path and use a Vmm_trie with a Map -- and store in the map a bit whether the unikernel is active or not -- here we can as well store the subscribers (and don't need anymore mutable state twice)

hannesm:
I think this is fine now. At least fine for review.

daemon/albatross_console.ml
```diff
@@ -14,9 +14,17 @@ open Lwt.Infix
...
-let active = ref Vmm_core.String_map.empty
+module LMap = Map.Make(Vmm_core.Name.Label)
 
-let read_console id name ring fd =
+(* The console output as a Vmm_ring.t with 1024 entries. The store is a trie,
+   where the key is the path, and the value is pair of "number of unikernels
+   allowed on this path" and "map", which key is the label, and its value a
+   triple: subscribers, active_or_not, ringbuffer. *)
+let state :
+  (int * ((Vmm_commands.version * Lwt_unix.file_descr) list * bool * string Vmm_ring.t) LMap.t) Vmm_trie.t ref =
```
reynir: Maybe worth it giving part of this type a name? And/or use a record? It's very long :D

daemon/albatross_console.ml
```diff
@@ -60,33 +68,37 @@ let read_console id name ring fd =
            the empty lines might have some semantic meaning *)
         let lines = List.filter (fun s -> not (String.equal "" s)) lines in
         Logs.debug (fun m -> m "read %u lines %u slack" (List.length lines) (String.length slack)) ;
-        let t = Ptime_clock.now () in
-        List.iter (fun line -> Vmm_ring.write ring (t, line)) lines;
+        let now = Ptime_clock.now () in
+        List.iter (fun line -> Vmm_ring.write ringbuffer (now, line)) lines;
+        let path = Vmm_core.Name.drop_label id in
+        let lbl = Option.get (Vmm_core.Name.name id) in
```
reynir: Do we ensure at call sites that it's a proper name?

daemon/albatross_console.ml
```diff
@@ -116,41 +128,86 @@ let open_fifo name =
...
+            match LMap.choose_opt (LMap.filter (fun _ (_, active, _) -> not active) map) with
+            | Some (key, (fds, _, _)) ->
+              Logs.warn (fun m -> m "dropping %s:%s"
+                            (Vmm_core.Name.to_string path)
+                            (Vmm_core.Name.Label.to_string key));
+              Lwt_list.iter_p Vmm_lwt.safe_close (List.map snd fds) >|= fun () ->
+              LMap.remove key map
+            | None ->
+              Logs.err (fun m -> m "intended to drop something while adding ringbuffer %s, but found nothing inactive" name);
+              Lwt.return map
```
reynir: I guess this can happen if we have a policy of N unikernels and later substitute the policy with N' < N unikernels. Or perhaps the "root" policy can override policies and add more unikernels? Can this become spammy in that case?

daemon/albatross_console.ml
```diff
@@ -116,41 +128,86 @@ let open_fifo name =
...
-    Lwt.async (fun () -> read_console id name ring f >|= fun () -> fifos `Close) ;
+    Lwt.async (fun () ->
+        read_console id name ringbuffer f >|= fun () ->
+        (match Vmm_trie.find path !state with
+         | None -> ()
+         | Some (_, map) ->
+           match LMap.find_opt lbl map with
+           | None -> ()
+           | Some (subs, _, ringbuffer) ->
+             let map = LMap.add lbl (subs, false, ringbuffer) map in
```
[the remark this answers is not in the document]
reynir: Oh, `read_console` returns when we for some reason can't read output from the unikernel. Ok, that's fine.

daemon/albatross_console.ml
reynir: We do since e74c0fe7fef0eedcd2b0bbdc0c4a0c4cc5368096

daemon/albatross_console.ml
hannesm: done in 9b3d2ac

PR #241, opened 2025-11-11
Add a new console command: Console_list_inactive, listing the inactive but available ring buffers

hannesm:
This is replied to with `Consoles of Name.t list.

This allows us to implement https://github.com/robur-coop/mollymawk/issues/126

### DIFF
--- a/daemon/albatross_console.ml
+++ b/daemon/albatross_console.ml
@@ -294,62 +294,49 @@ let handle max_subscribers s addr =
         let name = header.Vmm_commands.name in
         let path = Vmm_core.Name.path name in
         let lbl = Vmm_core.Name.name name in
-        match cmd  with
-        | `Console_add n ->
-          begin match lbl with
-            | None ->
-              (Vmm_lwt.write_wire s (header, `Failure "path used instead of name") >>= function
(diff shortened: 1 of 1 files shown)


daemon/albatross_console.ml
```diff
@@ -251,15 +251,16 @@ let handle s addr =
...
-            | Ok () -> Lwt.return_unit
-            | Error _ -> Vmm_lwt.safe_close s)
-        | path, Some name ->
-          match cmd with
-          | `Console_add n ->
-            begin
+        let path = Vmm_core.Name.path name in
+        let lbl = Vmm_core.Name.name name in
+        match cmd  with
+        | `Console_add n ->
```
reynir: We could match on `cmd, lbl` and then handle `lbl = None` for `` `Console_add _ | `Console_subscribe `` to reduce duplicated code a little. Then we would have a match `` `Console_list_inactive, _ ``

hannesm:
> just noticed that our close of the client fd behaviour is inconsistent and sometimes missing... esp. when the name is only a path, not a path -- here we have vmm_lwt.write calls and only in the error case a close.

addressed in #243

PR #243, opened 2025-11-11
albatross-console: close the client fds more appropriately:

hannesm:
- once the loop () in handle is done, the socket is closed --> remove some manual closes
- also avoid close in unsubscribe, do it at the call site if necessary

### DIFF
--- a/daemon/albatross_console.ml
+++ b/daemon/albatross_console.ml
@@ -51,7 +51,6 @@ type state = (int * console_map) Trie.t
 let state : state ref = ref Trie.empty
 
 let unsubscribe path lbl fd =
-  Vmm_lwt.safe_close fd >|= fun () ->
   match Trie.find path !state with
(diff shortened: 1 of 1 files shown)


daemon/albatross_console.ml
```diff
@@ -274,13 +274,18 @@ let send_history s version r (path, lbl) since =
...
+  Lwt_list.fold_left_s (fun s (i, v) ->
+      match s with
+      | None -> Lwt.return None
+      | Some s ->
+        let header = Vmm_commands.header ~version id in
+        let data = `Data (`Console_data (i, v)) in
+        Vmm_lwt.write_wire s (header, data) >|= function
+        | Ok () -> Some s
+        | Error _ ->
+          unsubscribe path lbl s;
```
reynir: So here we don't need to close the socket?

PR #246, opened 2025-12-01
Use Bstr.t instead of Cachet.bigstring

dinosaure:
We delete `Cachet.bigstring` on `cachet.0.0.3` and use `Bstr.t`.

### DIFF
--- a/albatross.opam
+++ b/albatross.opam
@@ -35,7 +35,8 @@ depends: [
   "http-lwt-client" {>= "0.3.0"}
   "happy-eyeballs-lwt"
   "solo5-elftool" {>= "0.4.0"}
-  "cachet" {>= "0.0.2"}
+  "cachet" {>= "0.0.3"}
+  "bstr"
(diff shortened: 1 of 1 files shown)


src/vmm_unix.ml
```diff
@@ -201,10 +201,10 @@ let destroy_tap tap =
 let cachet_of_str b =
   let map () ~pos len =
     if pos >= String.length b || len <= 0 then
-      (Cachet.Bstr.empty :> Cachet.bigstring)
+      (Cachet.Bstr.empty :> Bstr.t)
     else
       let len = min len (max 0 (String.length b - pos)) in
-      let pg : Cachet.bigstring = Bigarray.Array1.create Bigarray.char Bigarray.c_layout len in
```
dinosaure: We can also use `Bstr.create len` now if you want.

hannesm: where? in the line above, `let len = ...`?

reynir (approved):

Issue #219, opened 2025-06-13
feature request: separate the block handling (esp. create / dump / set) to a separate daemon


hannesm:
The reasoning is that this shouldn't require elevated privileges. Also, there may be a block provider using zfs features (while another one uses plain files, as now).

The albatross_daemon should still play a role, esp. since it tracks that a block device is used only once; and that a used block device is not destroyed. Plus of course that a create does not violate the resource policy.

FWIW, create should get rid of the optional data. This can be solved a layer above by doing a create followed by set.

Also, note that set and dump would be relaxed and possible if the block device is in use by a unikernel.

So, the idea is, similar to albatross_console, have a daemon where albatrossd talks to:
- create if passing the policy -> instruct the block daemon to create the device
- destroy if not used -> instruct block daemon to destroy it
- info would still be handled by albatrossd

and for the client, it directly communicates with block for:
- set - for writing stuff to the device
- dump - for retrieving stuff from the device

Any opinions? For the current functionality, no superuser privileges would be needed -- for a zfs block provider that uses snapshots etc., it'd need further privileges.

hannesm:
In #216 the block_add already has removed the optional data part. This also means the remaining things in this issue can be worked on, and shouldn't require any protocol changes.

PR #250, opened 2026-01-16
Be able to start and monitor BHyve virtual machines

hannesm:
This is not very well tested yet. EDIT: but a basic FreeBSD VM is starting :) [and console output is visible]

- [x] Test with a Linux VM (using the grub-bhyve path)
- [x] Test with multiple network interfaces
- [x] What is not done is pinning to specific CPU(s). This requires some further change(s), since there's only a single "cpu", but a BHyve VM may use more CPUs.
- [x] Also, the resources/resource usage needs to be adjusted for these BHyve VMs. At the moment, we have a short exit for /dev/zvol -- not albatross_daemon has a command-line argument `--allow-zvol-for` - as a temporary workaround (until #219 is a thing)
- [x] Only some common and standard things are supported, namely there's no support for `-d` for grub-bhyve. 
- [x] The restart-on-failure as well needs to be rethought / revised. <- this is fine, if a reboot is issued, the exit code of bhyve is 0, which matches the restart-on-failure. Exit code 1 is for "halt"

Later a path on how to install a virtual machine given you have nothing -- usually done via a boot CD or memstick... this is not very well thought out yet (but if you have an image, just block_set it, and you're done :)

### DIFF
--- a/src/vmm_asn.ml
+++ b/src/vmm_asn.ml
@@ -359,8 +359,9 @@ let v0_unikernel_config =
     and startup = None
     and fail_behaviour = `Quit (* TODO maybe set to restart by default :) *)
     and add_name = true
+    and cpuids = IS.singleton cpuid
     in
-    { typ ; compressed ; image ; fail_behaviour ; startup ; add_name ; cpuid ; memory ; block_devices ; bridges ; argv ; cpus = 1 ; linux_boot_partition = None }
+    { typ ; compressed ; image ; fail_behaviour ; startup ; add_name ; cpuids ; memory ; block_devices ; bridges ; argv ; cpus = 1 ; linux_boot_partition = None }
   and g _unikernel = failwith "cannot encode v0 unikernel configs"
   in
   Asn.S.map f g @@
@@ -381,8 +382,9 @@ let v1_unikernel_config =
     and block_devices = match blocks with None -> [] | Some xs -> List.map (fun b -> b, None, None) xs
     and startup = None
     and add_name = true
+    and cpuids = IS.singleton cpuid
     in
-    { typ ; compressed ; image ; fail_behaviour ; startup ; add_name ; cpuid ; memory ; block_devices ; bridges ; argv ; cpus = 1 ; linux_boot_partition = None }
+    { typ ; compressed ; image ; fail_behaviour ; startup ; add_name ; cpuids ; memory ; block_devices ; bridges ; argv ; cpus = 1 ; linux_boot_partition = None }
   and g _unikernel = failwith "cannot encode v1 unikernel configs"
   in
   Asn.S.(map f g @@ sequence @@
@@ -403,8 +405,9 @@ let v2_unikernel_config =
     and block_devices = match blocks with None -> [] | Some xs -> List.map (fun b -> b, None, None) xs
     and startup = None
     and add_name = true
+    and cpuids = IS.singleton cpuid
     in
-    { typ ; compressed ; image ; fail_behaviour ; startup ; add_name ; cpuid ; memory ; block_devices ; bridges ; argv ; cpus = 1 ; linux_boot_partition = None }
+    { typ ; compressed ; image ; fail_behaviour ; startup ; add_name ; cpuids ; memory ; block_devices ; bridges ; argv ; cpus = 1 ; linux_boot_partition = None }
   and g (unikernel : config) =
     let bridges =
       match unikernel.bridges with
@@ -413,8 +416,13 @@ let v2_unikernel_config =
     and blocks = match unikernel.block_devices with
       | [] -> None
       | xs -> Some (List.map (fun (a, _, _) -> a) xs)
+    and cpuid =
+      if IS.cardinal unikernel.cpuids <> 1 then
+        failwith "expected a single CPUid for v2 unikernel config"
+      else
+        IS.choose unikernel.cpuids
     in
-    (unikernel.typ, (unikernel.compressed, (unikernel.image, (unikernel.fail_behaviour, (unikernel.cpuid, (unikernel.memory, (blocks, (bridges, unikernel.argv))))))))
+    (unikernel.typ, (unikernel.compressed, (unikernel.image, (unikernel.fail_behaviour, (cpuid, (unikernel.memory, (blocks, (bridges, unikernel.argv))))))))
   in
   Asn.S.(map f g @@ sequence @@
            (required ~label:"typ" typ)
@@ -438,13 +446,19 @@ let v3_unikernel_config =
     and block_devices = match blocks with None -> [] | Some xs -> xs
     and startup = None
     and add_name = true
+    and cpuids = IS.singleton cpuid
     in
-    { typ ; compressed ; image ; fail_behaviour ; startup ; add_name ; cpuid ; memory ; block_devices ; bridges ; argv ; cpus = 1 ; linux_boot_partition = None }
+    { typ ; compressed ; image ; fail_behaviour ; startup ; add_name ; cpuids ; memory ; block_devices ; bridges ; argv ; cpus = 1 ; linux_boot_partition = None }
   and g (unikernel : config) =
     let bridges = match unikernel.bridges with [] -> None | xs -> Some xs
     and blocks = match unikernel.block_devices with [] -> None | xs -> Some xs
+    and cpuid =
+      if IS.cardinal unikernel.cpuids <> 1 then
+        failwith "expected a single CPUid for v3 unikernel config"
+      else
+        IS.choose unikernel.cpuids
     in
-    (unikernel.typ, (unikernel.compressed, (unikernel.image, (unikernel.fail_behaviour, (unikernel.cpuid, (unikernel.memory, (blocks, (bridges, unikernel.argv))))))))
+    (unikernel.typ, (unikernel.compressed, (unikernel.image, (unikernel.fail_behaviour, (cpuid, (unikernel.memory, (blocks, (bridges, unikernel.argv))))))))
   in
   Asn.S.(map f g @@ sequence @@
            (required ~label:"typ" typ)
@@ -467,19 +481,26 @@ let v3_unikernel_config =
                                 (optional ~label:"mac" mac_addr)))))
         -@ (optional ~label:"arguments"(my_explicit 2 (sequence_of utf8_string))))
 
-let unikernel_config =
+let v4_unikernel_config =
   let open Unikernel in
-  let f (typ, (compressed, (image, (startup, (add_name, (fail_behaviour, (cpuid, (memory, (blocks, (bridges, (argv, (cpus, linux_boot_partition)))))))))))) =
+  let f (typ, (compressed, (image, (startup, (add_name, (fail_behaviour, (cpuid, (memory, (blocks, (bridges, argv)))))))))) =
     let bridges = match bridges with None -> [] | Some xs -> xs
     and block_devices = match blocks with None -> [] | Some xs -> xs
-    and cpus = Option.value ~default:1 cpus
+    and cpus = 1
+    and cpuids = IS.singleton cpuid
+    and linux_boot_partition = None
     in
-    { typ ; compressed ; image ; fail_behaviour ; startup ; add_name ; cpuid ; memory ; block_devices ; bridges ; argv ; cpus ; linux_boot_partition }
+    { typ ; compressed ; image ; fail_behaviour ; startup ; add_name ; cpuids ; memory ; block_devices ; bridges ; argv ; cpus ; linux_boot_partition }
   and g (unikernel : config) =
     let bridges = match unikernel.bridges with [] -> None | xs -> Some xs
     and blocks = match unikernel.block_devices with [] -> None | xs -> Some xs
+    and cpuid =
+      if IS.cardinal unikernel.cpuids <> 1 then
+        failwith "expected a single CPUid for v4 unikernel config"
+      else
+        IS.choose unikernel.cpuids
     in
-    (unikernel.typ, (unikernel.compressed, (unikernel.image, (unikernel.startup, (unikernel.add_name, (unikernel.fail_behaviour, (unikernel.cpuid, (unikernel.memory, (blocks, (bridges, (unikernel.argv, (Some unikernel.cpus, unikernel.linux_boot_partition))))))))))))
+    (unikernel.typ, (unikernel.compressed, (unikernel.image, (unikernel.startup, (unikernel.add_name, (unikernel.fail_behaviour, (cpuid, (unikernel.memory, (blocks, (bridges, unikernel.argv))))))))))
   in
   Asn.S.(map f g @@ sequence @@
            (required ~label:"typ" typ)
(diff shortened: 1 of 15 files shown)


src/vmm_unix.ml
```diff
@@ -90,12 +90,12 @@ let read_fd_for_file = fd_for_file Unix.[ O_RDONLY ]
...
 let write_fd_for_file = fd_for_file Unix.[ O_WRONLY ; O_APPEND ]
 
-let null = match read_fd_for_file (Fpath.v "/dev/null") with
+let _null = match read_fd_for_file (Fpath.v "/dev/null") with
   | Ok fd -> fd
   | Error _ -> invalid_arg "cannot read /dev/null"
 
 let rec create_process prog args stdout =
-  try Unix.create_process prog args null stdout stdout with
+  try Unix.create_process prog args stdout stdout stdout with
```
hannesm, line 95: impact of this change is to be discussed

src/vmm_unix.ml
```diff
@@ -321,24 +321,77 @@ let bridges_exist bridges =
...
+        List.fold_left (fun n ((name, _, _) as arg) ->
+            let* n in
+            let block = bridge_name arg in
+            if String.equal name (string_of_int n) then
+              Ok (succ n)
+            else
+              Error (`Msg ("block device " ^ block ^ " with name " ^ name ^ " not in ascending order (expected " ^ string_of_int n ^ ")")))
+          (Ok 0) unikernel.block_devices
+      in
+      let name = "alba-" ^ string_of_int (Random.int 100_000) in
```
hannesm, line 388: Here, `Random.int` is used for identifying the VM. We could use mirage-crypto-rng -- should we? The issue with Random is that it may not be seeded (do we have to do that by hand? I always keep on forgetting).

hannesm: My observation is at least that the first VM is always named alba-9344 here on my laptop.

hannesm:
This is fine to review. I'd leave the CPU pinning out of this PR. It for sure requires some more testing, esp. with Linux guests.

src/vmm_unix.ml
```diff
@@ -321,24 +321,77 @@ let bridges_exist bridges =
...
+      let* () = prepare_bhyve name unikernel in
+      Ok name
```
hannesm, line 390: for bhyve, we encode the vmm name into the digest. This works fine, since the `digest` is the only means of communication, but I'm wondering whether it will break anything (that assumes digest be of a certain length, ...).

src/vmm_unix.ml
```diff
@@ -321,24 +321,77 @@ let bridges_exist bridges =
...
+            if String.equal name (string_of_int n) then
+              Ok (succ n)
+            else
+              Error (`Msg ("network device on " ^ bridge ^ " with name " ^ name ^ " not in ascending order (expected " ^ string_of_int n ^ ")")))
+          (Ok 0) unikernel.bridges
+      in
+      let* _ =
+        List.fold_left (fun n ((name, _, _) as arg) ->
+            let* n in
+            let block = bridge_name arg in
```
reynir: Here we use `bridge_name` for a block. Need to look at its definition...

src/vmm_unix.ml
reynir: We could use as many digits as needed to have the length the same as the digest (but then something might break because we use a dash and non-hex characters...)

src/vmm_unix.ml
reynir: I don't understand this change. So we use stdout for stdin?

reynir (commented):
Thanks! I wonder if we should guard this behind a `--experimental` flag? So we can sort of warn people that it might break or change in the near future?

About CPUs and CPU pinning. There doesn't have to be a one-to-one correspondence between cpu sets and numcpu - the OS can schedule the virtual CPUs on the allowed CPUs I suspect. So we could allow multiple `--cpu` arguments and for unikernels the numcpu is inherently fixed to 1.

About the digest: maybe we need to rethink the data structure so it can vary depending on the type (solo5/bhyve/...) with type appropriate fields. But I'm fine with this especially with an `--experimental` flag or so.

src/vmm_unix.ml
reynir: Ok, its type is `'a * 'a option * 'b -> 'a` and is fine to use. The name is slightly confusing here, so we can consider renaming it `device_name`.

src/vmm_unix.ml
hannesm: indeed, see 1d53d82e6ac0485a46705fb54f2326c0fbc2570c -- BHyve fails with /dev/null being the stdin. now since albatross_console never writes to the fifo, we're fine... :)

reynir: Could we use `fst (Unix.pipe ~cloexec:true ())` for bhyve? If so, maybe we make stdin an argument to `create_process` and pass either `null` for solo5 or `null_pipe` for bhyve assuming `let null_pipe = fst (Unix.pipe ~cloexec:true ())`? At this site we create a process for a program we don't know and the arguments are unknown as well. Edit: so it makes me a bit uneasy to pass such a fd here.

I couldn't figure out why `mevent_add(.., EVF_READ, ..)` fails for /dev/null or what it does.

hannesm:
(rebased & force-pushed)

src/vmm_unix.ml
hannesm: done in e50fdcd (a Random.self_init).

src/vmm_unix.ml
hannesm: since the digest is raw (in the unikernel case), we end up in the bhyve case that it is hex-encoded 'alba-<>'. I would prefer to break other apps there (i.e. early), and avoid having the length the same as other digests (which may be a more silent failure and a real failures later in the game^Wprogram).

src/vmm_unix.ml
hannesm: I renamed it to device_name in 8fc705a

src/vmm_unix.ml
hannesm: First of all, sory for my comment there. So, `mevent_add` is something bhyve-user-space specific. `mevent` is a "Micro event library for FreeBSD, based on kqueue". It basically puts a callback into kqueue via `kevent`.

Now, of course we can create some other resource (file descriptor) and pass it as `/dev/stdin`, we can as well have conditional code paths between BHyve and solo5-hvt. I'm happy to do this if you find reusing the already existing fifo troublesome.

I am at peace with using the fifo - the reason is that as I mentioned earlier the other side is albatross_console, and that never writes to the fifo. And BHyve only registers an event handler for reading there.

Let me know whether you'd like to still have a separate pipe here for stdin.

reynir: The file descriptor is as well not open for reading; only writing. I think it's okay. The function is not exported in the .mli, too.

reynir: I added a comment documenting this.

reynir:
The last two commits were meant to be one. It changes the printing so that the `digest` field is printed as "digest <hex>" for `` `Solo5 `` and "bhyve-vmname <digest>" for `` `BHyve ``. This should hopefully hide to the user that we abuse the same field for both.

hannesm:
That's excellent @reynir. I'm working on a patchset for having cpuids in the unikernel/vm context. this is a breaking change though... but it is fine to include here. :)

hannesm:
[52f59c2](https://github.com/robur-coop/albatross/pull/250/commits/52f59c2fbee96b8dae10543c64f630ab5b9aa947) revises the unikernel_info/config/arguments to carry a set of cpuids -- since this modifies the wire format in a non-backwards compatible way, I made the versioning dance...

This is as well supported for unikernels. since even there for a single-cpu unikernel it makes sense that this may be on cpu 4 and 8, so the hypervisor can schedule it depending on load of CPU 4 and 8...

hannesm:
CI error on ppc64 (I plan to ignore this):
```
(cd _build/default && /home/opam/.opam/5.4/bin/ocamlopt.opt -w @1..3@5..28@30..39@43@46..47@49..57@61..62-40 -strict-sequence -strict-formats -short-paths -keep-locs -g -I src/.albatross.objs/byte -I src/.albatross.objs/native -I /home/opam/.opam/5.4/lib/asn1-combinators -I /home/opam/.opam/5.4/lib/astring -I /home/opam/.opam/5.4/lib/bigstringaf -I /home/opam/.opam/5.4/lib/checkseum -I /home/opam/.opam/5.4/lib/decompress/de -I /home/opam/.opam/5.4/lib/decompress/zl -I /home/opam/.opam/5.4/lib/domain-name -I /home/opam/.opam/5.4/lib/duration -I /home/opam/.opam/5.4/lib/fmt -I /home/opam/.opam/5.4/lib/fpath -I /home/opam/.opam/5.4/lib/ipaddr -I /home/opam/.opam/5.4/lib/logs -I /home/opam/.opam/5.4/lib/macaddr -I /home/opam/.opam/5.4/lib/metrics -I /home/opam/.opam/5.4/lib/ohex -I /home/opam/.opam/5.4/lib/optint -I /home/opam/.opam/5.4/lib/ptime -cmi-file src/.albatross.objs/byte/vmm_asn.cmi -no-alias-deps -opaque -o src/.albatross.objs/native/vmm_asn.cmx -c -impl src/vmm_asn.ml)
/tmp/build_221c95_dune/camlasm270cda.s: Assembler messages:
/tmp/build_221c95_dune/camlasm270cda.s:8051: Error: operand out of range (0xfffffffffffead49 is not between 0xffffffffffff8000 and 0xffff)
```

So, somehow vmm_asn.ml isn't in good shape anymore for the assembler / ppc64 backend.

hannesm:
> Thanks! I wonder if we should guard this behind a --experimental flag? So we can sort of warn people that it might break or change in the near future?

I would like to not have such flags. The issue is that we need to care about backwards compatibility anyways in terms of what is stored on disk (or in certificates). So, an `--experimental` flag would in my opinion only lead to more conditionals... Also taking into account the number of deployed albatross instances on FreeBSD, it is likely only us needing to care about that. :)

src/vmm_asn.ml
```diff
@@ -862,14 +1054,15 @@ let unikernel_info =
...
     in
-    (unikernel.typ, (unikernel.startup, (unikernel.fail_behaviour, (unikernel.cpuid, (unikernel.memory, (unikernel.digest, (blocks, (bridges, (unikernel.argv, (Some unikernel.started, (Some unikernel.cpus, unikernel.linux_boot_partition)))))))))))
+    (unikernel.typ, (unikernel.startup, (unikernel.fail_behaviour, (cpuids, (unikernel.memory, (unikernel.digest, (blocks, (bridges, (unikernel.argv, (Some unikernel.started, (Some unikernel.cpus, unikernel.linux_boot_partition)))))))))))
   in
   Asn.S.(map f g @@ sequence @@
            (required ~label:"typ" typ)
          @ (optional ~label:"startup" int)
          @ (required ~label:"fail-behaviour" fail_behaviour)
-         @ (required ~label:"cpuid" int)
+         @ (required ~label:"cpuids" (set_of int))
```
reynir, line 1065: In the ASN.1 is it possible to make this
```OCaml
required ~label:"cpuids" (choice2 int (set_of int))
```
and have backwards compatibility?

src/vmm_core.ml
```diff
@@ -322,9 +322,10 @@ module Unikernel = struct
 
   let fine_with_policy (p : Policy.t) (c : config) =
     let bridge_allowed set s = String_set.mem s set in
-    if not (IS.mem c.cpuid p.cpuids) then
-      Error (`Msg (Fmt.str "CPUid of unikernel (%d) not allowed by policy (%a)"
-                     c.cpuid Fmt.(list ~sep:(any ", ") int) (IS.elements p.cpuids)))
+    if not (IS.subset c.cpuids p.cpuids) then
+      Error (`Msg (Fmt.str "CPUid of unikernel (%a) not allowed by policy (%a)"
+                     Fmt.(list ~sep:(any ", ") int) (IS.elements c.cpuids)
+                     Fmt.(list ~sep:(any ", ") int) (IS.elements p.cpuids)))
```
reynir: Here we could do set arithmetics to display what cpuids are not allowed. Can be done in a later PR.

src/vmm_core.ml
```diff
@@ -345,14 +346,15 @@ module Unikernel = struct
       Fmt.(option ((any "@") ++ Macaddr.pp)) mac
 
   let pp_config ppf (unikernel : config) =
-    Fmt.pf ppf "typ %a@ compression %B image %d bytes@ fail behaviour %a@ startup at %a@ add_name %B@ cpu %d@ %d MB memory@ block devices %a@ bridge %a@ cpus %u@ linux_boot %a"
+    Fmt.pf ppf "typ %a@ compression %B image %d bytes@ fail behaviour %a@ startup at %a@ add_name %B@ cpus %a@ %d MB memory@ block devices %a@ bridge %a@ cpus %u@ linux_boot %a"
```
reynir: now we twice have "cpus .." with different meaning. Maybe "cpuids" and "numcpu" or so?

src/vmm_core.ml
```diff
@@ -379,11 +381,12 @@ module Unikernel = struct
   }
 
   let pp_arguments ppf (unikernel : arguments) =
-    Fmt.pf ppf "fail behaviour %a@ startup at %a@ add_name %B@ cpu %d@ %d MB memory@ block devices %a@ bridge %a@ cpus %u@ linux_boot_partition %a"
+    Fmt.pf ppf "fail behaviour %a@ startup at %a@ add_name %B@ cpus %a@ %d MB memory@ block devices %a@ bridge %a@ cpus %u@ linux_boot_partition %a"
```
reynir: See comment further up.

src/vmm_core.ml
```diff
@@ -470,17 +473,18 @@ module Unikernel = struct
...
     in
     { typ = cfg.typ ; fail_behaviour = cfg.fail_behaviour ; startup = cfg.startup ;
-      cpuid = cfg.cpuid ; memory = cfg.memory ; block_devices ;
+      cpuids = cfg.cpuids ; memory = cfg.memory ; block_devices ;
       bridges ; argv = cfg.argv ; digest = t.digest ; started = t.started ;
       cpus = cfg.cpus ; linux_boot_partition = cfg.linux_boot_partition }
 
   let pp_info ppf (info : info) =
-    Fmt.pf ppf "typ %a@ started %a@ fail behaviour %a@ startup at %a@ cpu %u@ %u MB memory@ block devices %a@ bridge %a@ %a@ cpus %u@ linux_boot_partition %a"
+    Fmt.pf ppf "typ %a@ started %a@ fail behaviour %a@ startup at %a@ cpus %a@ %u MB memory@ block devices %a@ bridge %a@ %a@ cpus %u@ linux_boot_partition %a"
```
reynir: See comment further up.

src/vmm_resources.ml
```diff
@@ -167,8 +167,10 @@ let check_policy (p : Policy.t) (running_unikernels, used_memory) (unikernel : U
     Error (`Msg (Fmt.str
                    "maximum allowed memory (%d, used %d) would be exceeded (requesting %d)"
                    p.Policy.memory used_memory unikernel.Unikernel.memory))
-  else if not (IS.mem unikernel.Unikernel.cpuid p.Policy.cpuids) then
-    Error (`Msg (Fmt.str "CPUid %u is not allowed by policy" unikernel.Unikernel.cpuid))
+  else if not (IS.subset unikernel.Unikernel.cpuids p.Policy.cpuids) then
+    Error (`Msg (Fmt.str "CPUids %a is not allowed by policy (%a)"
+                   Fmt.(list ~sep:(any ", ") int) (IS.elements unikernel.cpuids)
+                   Fmt.(list ~sep:(any ", ") int) (IS.elements p.cpuids)))
```
reynir: Here as well we can subtract to get the disallowed CPU IDs.

reynir (commented):
The ASN.1 changes are quite large, but looks okay. I didn't review that part as thoroughly.

src/vmm_asn.ml
hannesm: Yes, sure. What I'm aiming for is "old clients should be able to talk to new servers". So, an old client data would then be accepted by the new server.

But, when the old client requests a unikernel_info, it would as well receive this new ASN with the set_of, and then fail to parse it.

Thus (and to keep unikernel_config and unikernel_info in sync), I prefer to change it as done here. The old client now sends unikernel_info, and the server decodes it as old_unikernel_info5 and replies with Old_unikernel_info5 -- in a form that the old client understands the layout and is able to deal with it.

Certainly, the other compatibility -- a new client talking to an old server -- is not given. Somehow I assume that servers are updated to the latest version ;)

src/vmm_core.ml
hannesm: done in 365f64e

src/vmm_resources.ml
hannesm: done in 365f64e

src/vmm_core.ml
hannesm: done in 8cf84d0

src/vmm_core.ml
hannesm: done in 8cf84d0

src/vmm_core.ml
hannesm: done in 8cf84d0

hannesm:
I went on the journey to test a Linux virtual machine, and it turns out that grub-bhyve is unhappy when called via Bos.Cmd. It does some output on stdout, and it also likes to get some input (or timeout).

Neither a pipe from `yes` nor from `echo "\n"` (or `echo -e "\n"`) worked (all leading to 100% CPU). What works is the Sys.command and waiting until the grub timeout is settled... I'm a bit clueless what to do here -- esp. since albatross treats only fully started or not started virtual machines (there's no concept of "configured, but not started") -- so the grub-bhyve / bhyveload (`prepare_ bhyve`) intermediate states don't exist.

Now, for grub this is slightly bad since it may result in an execution that never terminates (the grub-shell just waits for input). But that's maybe for another day / another PR.

hannesm:
so, I'll open some issues:
- how to _install_ a different vm
- run different vms on linux
- handle the grub-bhyve initialization in a nicer way (find a way to allow a user to push data there -- which basically requires standard input / a serial console -- and intermediate states

apart from that, I think we can merge and release.

hannesm:
Not sure what happened the day before yesterday. I've reverted the Sys.command change since it is not needed.

I also added a `timeout 1s` for bhyveload and grub-bhyve, so they will block the daemon for at most 1s.

Ready to merge and release, documentation still missing -- but he main outcome is: set `GRUB_TIMEOUT=0` in `/etc/default/grub` (and run `update-grub` afterwards) [at least that's how it works for debian].

PR #256, opened 2026-01-22
clean resources when anything after execution raises

### LINKED ISSUE #255: if there's an error between Vmm_unix.prepare and Vmm_unix.exec, resources are not (always) cleaned up
noticed when I had a code error in `insert_unikernel`. where `set_block_usage` raises an exception.

I'm not sure whether it is worth to actually fix this issue, since the `check_unikernel` just above should have verified that `insert_unikernel` will succeed...

hannesm:
fix #255

### DIFF
--- a/src/vmm_resources.ml
+++ b/src/vmm_resources.ml
@@ -97,15 +97,15 @@ let find_block t name = Vmm_trie.find name t.block_devices
 
 let set_block_usage t name active =
   match Vmm_trie.find name t with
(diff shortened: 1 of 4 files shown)


hannesm:
I force-pushed a different solution, and removed the exceptions from Vmm_resources.insert_unikernel. This way if the `success` branch of the create continuation fails, the `fail` continuation is called, which cleans up resources.

reynir (approved):
Thanks.

PR #257, opened 2026-01-22
Also dump the state if we're coming out of a restart handler, and will not restart a unikernel

hannesm:
If we have a unikernel that exited on a non-restartable exit code, we should dump the state to not restart it upon next albatross start. Fixup for observed #249 fallout.


reynir (approved):
Thanks!

PR #261, opened 2026-03-20
Add a replace-policy sub-command alias

### LINKED ISSUE #260: Modify policy
@dinosaure asked if it was possible to modify a policy, and it's only possible to remove and add policies at the moment.

reynir:
And document the behavior of

    albatross-client {add-policy|replace-policy}

a bit better.

Closes #260. CC @dinosaure


hannesm:
thanks

PR #262, opened 2026-04-11
when reading console data, before pushing it to the ring buffer, escape it

hannesm:
The reason is that it may be arbitrary binary data, and it will be forwarded to a potential albatross-client, and may mess up your console (or whatever other device)

//cc @reynir

### DIFF
--- a/daemon/albatross_console.ml
+++ b/daemon/albatross_console.ml
@@ -83,7 +83,7 @@ let read_console (path, lbl) name ringbuffer fd =
         let get_slack lines =
             match List.rev lines with
             | [] | [ _ ] -> lines, ""
-            | slack::tl -> List.rev tl, slack
+            | slack::tl -> List.map String.escaped (List.rev tl), slack
         in
         let mode, lines, slack =
           (* anything exceeding 512 bytes get dropped until a newline is read *)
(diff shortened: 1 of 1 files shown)


reynir:
What do you think about pushing the escaping to the client? Different clients (albatross-client, mollymawk) may want to escape differently.

daemon/albatross_console.ml
```diff
@@ -104,10 +104,11 @@ let read_console (path, lbl) name ringbuffer fd =
             match String.split_on_char '\n' data with
             | [] -> assert false
             | [ x ] ->
-              if String.length x >= 512 then
+              let esc = String.escaped x in
+              if String.length esc >= 512 then
                 (* if there is no newline, drop *)
                 let data =
-                  if String.length x = 512 then x else String.sub x 0 512
```
reynir, line 110: Maybe escape *after* cut off? For example if there is a carriage return at byte 511 then we might get a string that ends with a backslash (`...\`) which itself doesn't have any meaning when the string is escaped.

reynir:
A downside i see is that `String.escaped` is a bit eager to escape unicode. For a cmdliner manual page this can result in this (ocaml toplevel session):

```OCaml
# String.escaped "       traceroute [OPTION]…";;
- : string = "       traceroute [OPTION]\\226\\128\\166"
# String.escaped "       traceroute [OPTION]…" |> print_endline;;
       traceroute [OPTION]\226\128\166
```

If there is an immediate concern I'm fine with escaping and then we can reiterate later on.

hannesm:
superseeded by #266

PR #269, opened 2026-05-05
Ensure on a "Destroy" that the state is dumped, fixes #268

### LINKED ISSUE #268: State is not dumped when needed
I have various fast-restarting unikernels on my laptop. When I `albatross-client destroy` them, they're gone. But the state file on disk is not updated, this means when I reboot they're again running.

This is related to #249 (and #257) -- previously the `Vmm_vmmd.waiter` did the `dump_state`. Now, it is done slightly differently in the callers thereof if needed.

hannesm:


hannesm:
I tested this, and it works nicely. I'll merge and cut a release, since the current shape isn't very nice... the destroyed unikernels will wakeup after reboot / albatross daemon restart

PR #271, opened 2026-05-06
Allow using --block and --network in albatross-client add-policy

reynir:
It is closer to the names we use in `albatross-client create`


hannesm:
well observed and nicely done

PR #276, opened 2026-06-18
restart: respect memory ; numcpus ; linux_boot_partition

### LINKED ISSUE #275: Restart with arguments doesn't change memory
When we restart with arguments and change memory value, this gets ignored. Is this intentional?

cc @hannesm

hannesm:
fixes #275 /cc @PizieDust


PizieDust (approved):
Thank you @hannesm

PR #280, opened 2026-07-05
add commands for albatross logging service

### LINKED ISSUE #278: Reintroduce albatross-logs
`albatross-logs` were removed sometime in this commit https://github.com/robur-coop/albatross/pull/75.
We now have a real use-case for albatross emitting events, as such a suitable client, mollymawk, can greatly benefit from them. 

For example, while working on autoscaling, in this PR: https://github.com/robur-coop/mollymawk/pull/253, we have a background task that runs every 600 seconds with the goal of trying to find clean up the state machine of vms and their clones if it gets dirty over time. While this works, having long running background task is not a good design. Rather, albatross can just emit an event when a vm is destroyed and mollymawk which is listening, updates its logbook. 

Another good case is, when mollymawk scales a vm and creates a clone, the albatross create event will signal when to update the load-balancer, and the same when a clone is destroyed.

With mollymawk being able to support multiple albatross servers, knowing when servers are active or dead can be helpful as well.

In terms of reporting, if a primary vm stops abruptly, the emitted event can allow mollymawk be able to send an email to the concerned user with maybe a link to the last logs captured.

Off the top of my head, I think the following events will be helpful:

- `Unikernel_start (name, digest, pid, taps, blocks)`
- `Unikernel_stop (name, pid, status)`
- `Startup` to detect if albatross has started
- `Disconnected` to detect if albatross has gone offline

we can as well think of events for block devices and network interfaces. 

Will `console_logs` still exists with `albatross-logs`?

cc @hannesm

hannesm:
The log service (part of vmm_vmmd / albatrossd) acts as follows:
- every path can subscribe once (existing subscriber is kicked)
- whenever a unikernel (below that subscribed path) is started or stopped, the subscriber receives a log event

Client: add log subcommand

fixes #278 //cc @PizieDust please tell which events and which data carried in the events would be useful -- also whether the "one log sink per path" is fine (I guess similar to stats, there's only the need from mollymawk to subscribe once to the log).

### DIFF
--- a/client/albatross_client.ml
+++ b/client/albatross_client.ml
@@ -849,6 +849,9 @@ let console () since count =
 let inactive_consoles () path =
   jump (`Console_cmd `Console_list_inactive) (Vmm_core.Name.make_of_path path)
 
+let log_subscribe () path =
+  jump (`Log_cmd `Log_subscribe) (Vmm_core.Name.make_of_path path)
+
 let stats_add () vmmdev pid bridge_taps =
   let vmmdev = Option.value ~default:"" vmmdev in
   jump (`Stats_cmd (`Stats_add (vmmdev, pid, bridge_taps)))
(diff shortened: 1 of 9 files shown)


hannesm:
So, @PizieDust, would this as a start be good for mollymawk, or are there other events you're interested in?

I'm slightly undecided about whether to include the `Name.t` in the `Logging.t` variant or not -- given that on the wire, the header already contains the name.

PizieDust:
> So, @PizieDust, would this as a start be good for mollymawk, or are there other events you're interested in?
> 
> I'm slightly undecided about whether to include the `Name.t` in the `Logging.t` variant or not -- given that on the wire, the header already contains the name.

Thanks for your work on this. I don't think we need to pass the Name.t as we already have it in the header.

PR #283, opened 2026-07-14
Stats rework

hannesm:
First step: remove bhyve debug counters.


hannesm:
What we backward-oncompatible loose is the Stats_add -- but since this is daemon-only, this is fine (as they are updated in one go via the packaging).

PizieDust:
I've tested this. I opened a PR in mollymawk: https://github.com/robur-coop/mollymawk/pull/266

PizieDust (approved):
Thanks @hannesm 
This works fine for me.

Is it worth it to publish a little article about no longer having `albatross_influx` daemon?

hannesm:
I tested this locally, and it works nicely. I cut a release.

PR #284, opened 2026-08-04
update README: other VMs, no bhyve statistics, no separate influx daemon

hannesm:


hannesm:
CI failure is unrelated

PR #291, opened 2026-09-16
Use getprotobyname to instantiate a TCP/IP socket instead of relying on the system

dinosaure:
On my `man 2 socket`, I have this description:

> The protocol specifies a particular protocol to be used with the socket.  Normally only a single protocol exists to support a  particular socket type within a given protocol family, in which case protocol can be specified as 0.  However, it is possible that many protocols  may  exist,  in which case a particular protocol must be specified in this manner.  The protocol number to use is specific to the “communication domain” in which communication is to take place; see protocols(5).  See getprotoent(3)  on  how  to map  protocol  name strings to protocol numbers.

And I can see on my `/etc/protocols`:
```
ip	0	IP		# internet protocol, pseudo protocol number
```

So when I try to start `albatross_stats` on my server, I fail with this error:
```
socket(AF_INET, SOCK_STREAM, IPPROTO_IP) = -1 EAFNOSUPPORT (Address family not supported by protocol)
```

This patch ensures that we allocates a TCP/IP socket.


src/vmm_lwt.ml
```diff
@@ -47,7 +48,8 @@ let service_socket sock =
           | e -> raise e)
     | false -> Lwt.return_unit)
   >>= fun () ->
-  let s = Lwt_unix.(socket PF_UNIX SOCK_STREAM 0) in
+  let proto = Unix.getprotobyname "tcp" in
```
hannesm, line 51: that's not correct. this is a unix domain socket here.

src/vmm_lwt.ml
```diff
@@ -62,7 +64,8 @@ let service_socket sock =
   s
 
 let connect addrtype sockaddr =
-  let c = Lwt_unix.(socket addrtype SOCK_STREAM 0) in
+  let proto = Unix.getprotobyname "tcp" in
```
hannesm, line 67: this is not correct -- depending on the passed `addrtype` this is either PF_UNIX or PF_INET/PF_INET6 -- and only in the latter case we're using TCP.

src/vmm_lwt.ml
dinosaure: Yes, sorry I did something really mechanical.

hannesm:
I guess with #292 this is superseeded?

hannesm:
not needed

PR #292, opened 2026-09-16
Allow AF_INET for albatross_stats.daemon

dinosaure:
Currently, `albatross_stats.daemon` would like to start a `AF_INET` connection when we would like to communicate with `influx`/`telegraf`. We allow such connection on the `systemd` level.


hannesm:
thanks

